Issue #22, opened 2018-06-05
Analytics module


tagantroy:
Some extra metrics required for "smart" batching and sorting.

- [ ] Save execution results to InfluxDb
- [ ] Query test success rate
- [ ] Query test execution time(percentile)

PR #23, opened 2018-06-05
[WIP] Upload metrics to InfluxDb

tagantroy:
#22 We need store some metrics for sorting, flakiness, batching strategies. 

- [x] Integrate InfluxDb client
- [x] Implement tracker
- [x] Implement metrics provider 

revert before merge:
- [x] Deployment.kt changes
- [x] Configuration.kt remove debug code and uncomment line 87


core/src/main/kotlin/com/malinskiy/marathon/Marathon.kt
```diff
@@ -1,11 +1,11 @@
 package com.malinskiy.marathon
 
 import com.google.gson.Gson
-import com.malinskiy.marathon.analytics.DelegatingTracker
-import com.malinskiy.marathon.analytics.local.DeviceTracker
-import com.malinskiy.marathon.analytics.local.JUnitTracker
-import com.malinskiy.marathon.analytics.Tracker
-import com.malinskiy.marathon.analytics.local.TestRusultsTracker
+import com.malinskiy.marathon.analytics.*
```
dsvoronin: we banned wildcard imports in out projects just to forget that it could be a problem
IDE's can collapse it by default, so verboseness not a problem either
what's your opinion?

p.s. detekt should handle it by simple rule afaik

core/build.gradle.kts
```diff
@@ -25,6 +25,7 @@ dependencies {
     implementation(Libraries.kotlinStdLib)
     implementation(Libraries.kotlinCoroutines)
     implementation(Libraries.kotlinLogging)
+    implementation(Libraries.influxDbClient)
```
dsvoronin, line 27: maybe it will be better to create separate module with influx implementation and leave only interfaces in `:core` ?

core/src/main/kotlin/com/malinskiy/marathon/execution/Configuration.kt
```diff
@@ -81,6 +84,9 @@ data class Configuration constructor(
                     outputDir = outputDir,
                     applicationOutput = applicationOutput,
                     testApplicationOutput = testApplicationOutput,
+//                    analyticsConfiguration = analyticsConfiguration ?: AnalyticsConfiguration.DisabledAnalytics,
```
dsvoronin: i'm not a fan of commented code in master branch

dsvoronin (approved):

core/src/main/kotlin/com/malinskiy/marathon/execution/Configuration.kt
tagantroy: yes, sure. I'll remove it before merge

core/build.gradle.kts
tagantroy: Yes, it is better to create separate module and load it with ServiceLoader

Issue #34, opened 2018-06-27
Filtering of executed tests


Malinskiy:
Whitelist + blacklist of tests by
- [x]  test class simple or FQN name
- [x]  test package
- [x]  FQN test annotation

The logic should get all the available tests first, then it should apply all the whitelists from different groups (if available), then apply all blacklists to everything that is still left (if available)

tagantroy:
merged to develop. 0.1.3 release

PR #41, opened 2018-07-04
Feature/test filtering

Malinskiy:
Implements #34


core/src/main/kotlin/com/malinskiy/marathon/Marathon.kt
```diff
@@ -3,9 +3,7 @@ package com.malinskiy.marathon
 import com.google.gson.Gson
 import com.malinskiy.marathon.analytics.AnalyticsFactory
 import com.malinskiy.marathon.device.DeviceProvider
-import com.malinskiy.marathon.execution.Configuration
-import com.malinskiy.marathon.execution.Scheduler
-import com.malinskiy.marathon.execution.TestParser
+import com.malinskiy.marathon.execution.*
```
tagantroy: use explicit imports instead of wildcard

tagantroy (commented):
Can you execute detektCheck and fix warnings?

tagantroy (approved):

PR #71, opened 2018-07-18
Fix branches to build on github

Malinskiy:


Malinskiy:
I've added the github token in Travis, let's see how the next release goes with this

PR #100, opened 2018-09-07
Refactor Trackers to support raw data from execution of tests

Malinskiy:
- Tracker now has trackTestTransition method to log all test transitions. For backwards compatibility NoOpTracker backports trackTestFinished()
- Housekeeping (detekt issues)


core/src/main/kotlin/com/malinskiy/marathon/execution/queue/TestResultReporter.kt
```diff
@@ -58,24 +59,13 @@ class TestResultReporter(private val poolId: DevicePoolId,
             }
         }
         onTransition {
-            val validTransition = it as? StateMachine.Transition.Valid
-            if (validTransition !is StateMachine.Transition.Valid) {
+            if (it as? StateMachine.Transition.Valid !is StateMachine.Transition.Valid) {
                 logger.error { "from ${it.fromState} event ${it.event}" }
-                return@onTransition
```
tagantroy, line 64: I hope we don't have invalid transactions in current implementation otherwise analytics will track it too.

tagantroy (approved):

cdsap (approved):

Issue #111, opened 2018-10-17
Support test running for library projects


lukaville:
Currently `testApplicationOutput` param is required and `marathon-gradle-plugin` throws an exception if you try to apply it to a library project:

```
A problem occurred configuring project ':components:SomeUiPart'.
> com.android.build.gradle.internal.api.LibraryVariantOutputImpl_Decorated cannot be cast to com.android.build.gradle.api.ApkVariantOutput
```

Malinskiy:
@lukaville How do you plan to test UI without the support from the framework? Unless I'm mistaken you can't test UI [without an application](https://developer.android.com/training/testing/espresso/basics)

lukaville:
@Malinskiy You can run UI test for library project in the same way like for app project.

If you run `:some-library:assembleDebugAndroidTest` task it will assemble test apk with **library** itself and with **tests**: `some-library-debug-androidTest.apk`.

Then if you install the apk you can run the tests like this: `adb shell am instrument -w -r -e debug false -e class 'com.library.SomeTest' com.library.test/android.support.test.runner.AndroidJUnitRunner`. By default, I think it will use `android.app.Application` but it's customizable by using manifest or custom runner.

rasskazovaleksey:
Espresso can't but it's possible with uiAutomator. Would be a nice feature

lukaville:
Espresso works fine with library projects, so it's not a problem

tagantroy:
Fixed #112

PR #112, opened 2018-10-26
Support android library projects

tagantroy:
Fix for #111 
- [x] Add sample android library project
- [x] Gradle Plugin: Support library project
- [x] CLI: Support library project


cdsap (approved):
LGTM, only comments about versions but not important for the PR

PR #124, opened 2018-11-09
Fixing multiple issues related to test runs

Malinskiy:
List of fixes:
- Sometimes device termination leads to active batch present. First terminate the device then return batch
- In order to easily dogfood simplify custom deployments
- Logging was not present in gradle plugin case. Workaround by outputting to warn level
- Add more logs to some steps
- AndroidDevice provider shouldn't create another instance if a device with the same serial was already observed
- Timeline should now have proper test names, include test retires to better understand what happened during the run
- ExecutionTimeSortingStrategy now has 1 test :)
- Fix bug with retention policy not applied to all InfluxDB instances

### DIFF
--- a/core/src/main/kotlin/com/malinskiy/marathon/analytics/Analytics.kt
+++ b/core/src/main/kotlin/com/malinskiy/marathon/analytics/Analytics.kt
@@ -3,4 +3,4 @@ package com.malinskiy.marathon.analytics
 import com.malinskiy.marathon.analytics.metrics.MetricsProvider
 import com.malinskiy.marathon.analytics.tracker.Tracker
 
-class Analytics(tracker: Tracker, metricsProvider: MetricsProvider) : Tracker by tracker, MetricsProvider by metricsProvider
+class Analytics(tracker: Tracker, val metricsProvider: MetricsProvider) : Tracker by tracker, MetricsProvider by metricsProvider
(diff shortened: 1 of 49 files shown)


tagantroy (approved):

core/src/main/kotlin/com/malinskiy/marathon/execution/strategy/impl/batching/FixedSizeBatchingStrategy.kt
```diff
@@ -19,6 +32,15 @@ class FixedSizeBatchingStrategy(@JsonProperty("size") private val size: Int) : B
             } else {
                 result.add(item)
             }
+
+            if(durationMillis != null && percentile != null && timeLimit != null) {
+                //Check for expected batch duration. If we hit the duration limit - break
+                //Important part is to add at least one test so that if one test is longer than a batch
+                //We still have at least one test
+                val expectedTestDuration = analytics.metricsProvider.executionTime(item, percentile, timeLimit)
```
tagantroy, line 40: Why not ```analytics.executionTime(...)```?

core/src/main/kotlin/com/malinskiy/marathon/analytics/Analytics.kt
```diff
@@ -3,4 +3,4 @@ package com.malinskiy.marathon.analytics
 import com.malinskiy.marathon.analytics.metrics.MetricsProvider
 import com.malinskiy.marathon.analytics.tracker.Tracker
 
-class Analytics(tracker: Tracker, metricsProvider: MetricsProvider) : Tracker by tracker, MetricsProvider by metricsProvider
+class Analytics(tracker: Tracker, val metricsProvider: MetricsProvider) : Tracker by tracker, MetricsProvider by metricsProvider
```
tagantroy, line 6: I think we don't need expose MetricsProvider because Analytics delegates work to metricsProvider

PR #173, opened 2019-02-14
Fix progress calculation

### LINKED ISSUE #134: Progress report is broken (shows more than 100%)
Some users are seeing progress reporter sometimes showing progress of more than 100%. I've also personally have seen progress of -3%.

Suspected bug is in handling of the ```INCOMPLETE``` tests, they shouldn't count towards the execution.

Malinskiy:
Fixed #134 

This also introduces the usual Microsoft problem of progress bars, that is progress can show 97% -> 100% -> 97% -> 100%

There is nothing we can do here if we don't change the logic of calculating progress including the retries.

### DIFF
--- a/core/src/main/kotlin/com/malinskiy/marathon/execution/progress/tracker/PoolProgressTracker.kt
+++ b/core/src/main/kotlin/com/malinskiy/marathon/execution/progress/tracker/PoolProgressTracker.kt
@@ -42,33 +42,36 @@ class PoolProgressTracker {
         }
     }
 
-    private fun updateStatus(test: Test, newStatus: ProgressEvent) {
-        tests[test]?.transition(newStatus)
-    }
+    private fun updateStatus(test: Test, newStatus: ProgressEvent) = tests[test]?.transition(newStatus)
 
     private val totalTests = AtomicInteger(0)
     private val completed = AtomicInteger(0)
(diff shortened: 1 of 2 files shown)


tagantroy (approved):

core/src/main/kotlin/com/malinskiy/marathon/execution/progress/tracker/PoolProgressTracker.kt
```diff
@@ -42,33 +42,36 @@ class PoolProgressTracker {
...
         updateStatus(test, ProgressEvent.Failed)
         failed.updateAndGet {
             it + 1
         }
     }
 
-    fun testPassed(test: Test, device: DeviceInfo) {
+    fun testPassed(test: Test) {
         completed.updateAndGet {
             it + 1
```
memega, line 69: You might want to add an early exit here as well to account for multiple successful executions

Malinskiy: This will not work as expected since multiple success executions of tests are actually not affecting our percentage

Malinskiy:
There is still an known issue with the parameterized tests since we can't check how many do we have when doing the dex parsing

PR #181, opened 2019-02-19
[vendor-android] Implement clearing application and test application data

### LINKED ISSUE #178: Add support for clearing package data between batches for Android
One of the critical features for teams that are struggling with sharing state between test runs is clearing the local files for application. Orchestrator [supports](https://github.com/android/android-test/blob/master/runner/android_test_orchestrator/java/androidx/test/orchestrator/AndroidTestOrchestrator.java#L352) this as parameter. We should implement a similar option that clears the package data before each batch is executed. This will introduce some additional time for testing since this command takes some time, but it's better than implementing support for orchestrator and introducing a new dependency.

Malinskiy:
Clearing happens before executing each test batch.

Controlled via *applicationPmClear* and *testApplicationPmClear* parameters in Marathonfile and gradle plugin

Fixes #178


vendor/vendor-android/src/main/kotlin/com/malinskiy/marathon/android/DdmlibDeviceExtensions.kt
```diff
@@ -94,4 +116,17 @@ fun getScreenRecorderCommand(@NonNull remoteFilePath: String,
     sb.append(remoteFilePath)
 
     return sb.toString()
-}
\ No newline at end of file
+}
+
+class SimpleOutputReceiver: MultiLineReceiver() {
+    private val buffer = StringBuffer()
```
tagantroy, line 122: Do we need StringBuffer here or it can be replaced with StringBuilder? StringBuilder is not safe for multithreaded usage. I'm not sure, but probably this received will not be used from multiple threads

tagantroy (approved):

vendor/vendor-android/src/main/kotlin/com/malinskiy/marathon/android/DdmlibDeviceExtensions.kt
Malinskiy: I'm not sure. Supposedly Ddmlib should call it from single thread, but I don't trust it

PR #190, opened 2019-03-14
Feature/Add preferable recorder type config

Tieru:
Currently, if a device allows capturing both videos and gifs, marathon always chooses video. If it fails to capture video, there is no way to request gifs records. With this config, we can select how we would like to capture test progress.

Any feedback is welcome :)

### DIFF
--- a/cli/src/test/resources/fixture/config/sample_1.yaml
+++ b/cli/src/test/resources/fixture/config/sample_1.yaml
@@ -70,3 +70,4 @@ vendorConfiguration:
   testApplicationPmClear: true
   instrumentationArgs:
     debug: "false"
+  preferableRecorderType: "screenshot"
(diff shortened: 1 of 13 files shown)


tagantroy (approved):

cli/src/test/resources/fixture/config/sample_1.yaml
```diff
@@ -69,4 +69,5 @@ vendorConfiguration:
   applicationPmClear: true
   testApplicationPmClear: true
   instrumentationArgs:
-    debug: "false"
\ No newline at end of file
+    debug: "false"
+  preferableRecorderType: "screenshot"
```
tagantroy, line 73: 👍

tagantroy (approved):
Good job! Thanks for contribution.

PR #191, opened 2019-03-15
Add install options configuration

lukaville:
`-r` flag doesn't allow to force install the apk in case of `INSTALL_FAILED_VERSION_DOWNGRADE` error, it is necessary to pass `-d` param for this. Therefore I've added optional installOptions param.


tagantroy (changes requested):
Please add installOptions to cli configuration

tagantroy (changes requested):
Hello, Please add check to ConfigFactorySpec.kt.
Add installOptions to https://github.com/Malinskiy/marathon/blob/7c62c435c511465ce699dd1630214498a24a5b66/cli/src/test/resources/fixture/config/sample_1.yaml and modify ConfigFactorySpec.kt. 
Thank you in advance.

tagantroy (approved):

Issue #199, opened 2019-03-22
Marathon tries to execute "#null" test


DSamaryan:
I faced a strange behaviour when marathon tried to execute "#null" test, it printed the following logs:
```
22% | [omni]-[EMULATOR28X0X23X0] com.android.kakao.ClustersTest#choiceTest started
10:11:39.121 [AndroidDevice - execution - emulator-5554 @coroutine#148] INFO DebugTestRunListener - testEnded EMULATOR28X0X23X0 test = com.android.kakao.ClustersTest#choiceTest
23% | [omni]-[EMULATOR28X0X23X0] com.android.kakao.ClustersTest#choiceTest ended
10:11:39.650 [AndroidDevice - execution - emulator-5554 @coroutine#148] INFO DebugTestRunListener - testRunEnded elapsedTime 30130
10:11:39.651 [main @coroutine#6] DEBUG QueueActor[DevicePoolId(name=omni)] - handle test results EMULATOR28X0X23X0
10:11:39.654 [main @coroutine#6] DEBUG QueueActor[DevicePoolId(name=omni)] - request next batch for device EMULATOR28X0X23X0
10:11:39.655 [main @coroutine#6] DEBUG QueueActor[DevicePoolId(name=omni)] - sending next batch for device EMULATOR28X0X23X0
10:11:39.655 [main @coroutine#7] DEBUG DevicePool[omni]_DeviceActor[EMULATOR28X0X23X0] - executeBatch EMULATOR28X0X23X0
10:11:39.656 [AndroidDevice - execution - emulator-5554 @coroutine#152] DEBUG AndroidDeviceTestRunner - tests = [com.android.kakao.ClustersTest#choiceCityCluster]
10:11:49.534 [AndroidDevice - execution - emulator-5554 @coroutine#152] INFO DebugTestRunListener - testRunStarted EMULATOR28X0X23X0
10:11:49.534 [AndroidDevice - execution - emulator-5554 @coroutine#152] INFO DebugTestRunListener - testStarted EMULATOR28X0X23X0 test = com.android.kakao.ClustersTest#null
23% | [omni]-[EMULATOR28X0X23X0] com.android.kakao.ClustersTest#null started
[...]

PR #200, opened 2019-03-29
Ignored test results from ddm lib with null method name

xanderblinov:
Ignored test results from ddm lib with null method name. 
Fix for issue https://github.com/Malinskiy/marathon/issues/199


tagantroy:
Hello @xanderblinov. Please add a test for this changes in sample app. Thanks in advance.

PR #205, opened 2019-04-09
Use ddmlib api to obtain external storage directory

tagantroy:


vendor/vendor-android/src/main/kotlin/com/malinskiy/marathon/android/RemoteFileManager.kt
```diff
@@ -10,28 +11,33 @@ import com.malinskiy.marathon.log.MarathonLogging
 
 import java.io.IOException
 
-object RemoteFileManager {
+class RemoteFileManager(private val device: IDevice) {
```
Malinskiy, line 14: Since we're now passing the device in the constructor and it becomes one instance of RemoteFileManager per device it doesn't make sense that other methods like removeRemotePath still have a device reference as a parameter because it's redundant.

Either cache the outputDir per device (create a Map from IDevice to outputDir) and still have an object or modify this manager to basically not request IDevice in other methods.

Malinskiy (changes requested):

vendor/vendor-android/src/main/kotlin/com/malinskiy/marathon/android/RemoteFileManager.kt
tagantroy: Oh, yes. Let me remove IDevice from params

tagantroy: Done.

Issue #133, opened 2018-12-13
Video recorder is broken


tagantroy:
Some of the videos are broken. It looks like a problem with headers(I checked in hex editor).

Malinskiy:
@tagantroy can you please provide a sample of the file?

tagantroy:
Reproduced on latest develop.
[com.example.MainActivityFlakyTest#testTextFlaky3.mp4.zip](https://github.com/Malinskiy/marathon/files/2986476/com.example.MainActivityFlakyTest.testTextFlaky3.mp4.zip)


It looks like a problem with headers

PR #207, opened 2019-04-09
Screenrecord workaround

tagantroy:
Workaround for #133

### DIFF
--- a/vendor/vendor-android/src/main/kotlin/com/malinskiy/marathon/android/executor/listeners/video/ScreenRecorderStopper.kt
+++ b/vendor/vendor-android/src/main/kotlin/com/malinskiy/marathon/android/executor/listeners/video/ScreenRecorderStopper.kt
@@ -17,17 +17,26 @@ internal class ScreenRecorderStopper(private val deviceInterface: IDevice) {
         }
     }
 
+    private fun grepPid(receiver: CollectingShellOutputReceiver) {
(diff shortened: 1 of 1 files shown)


vendor/vendor-android/src/main/kotlin/com/malinskiy/marathon/android/executor/listeners/video/ScreenRecorderStopper.kt
```diff
@@ -55,8 +64,8 @@ internal class ScreenRecorderStopper(private val deviceInterface: IDevice) {
 
     companion object {
         private val logger = MarathonLogging.logger("ScreenRecorderStopper")
-        private const val SCREEN_RECORD_KILL_ATTEMPTS = 100
-        private const val PAUSE_BETWEEN_RECORDER_PROCESS_KILL = 50
+        private const val SCREEN_RECORD_KILL_ATTEMPTS = 5
+        private const val PAUSE_BETWEEN_RECORDER_PROCESS_KILL = 300
```
Malinskiy, line 71: Can you please mention as a comment in the code that we've empirically found this to be the only way to fix the broken video files? These delays of random millis are really confusing otherwise. And a link to the issue would also be nice.

Malinskiy (changes requested):

Malinskiy (approved):

Issue #185, opened 2019-02-25
InfluxMetricsProvider does not update query results when parameters change


memega:
After `InfluxMetricsProvider.executionTime` is called first time, query result is permanently cached and will be used in subsequent calls even with different `percentile` and `limit` values. It is problematic in current implementation because both sorting and batching strategies use the same instance of metrics provider. Sorting strategy normally initiates the first request. When batching strategy requests data later, it would be cached.

Malinskiy:
This is fixed by @memega in his branch, let's merge it

PR #216, opened 2019-04-13
InfluxMetricsProvider cache multiple sets of data

tagantroy:
Fix for #185. 
Thanks @memega 

1. Remove unused import
2. Fix code format

### DIFF
--- a/core/src/main/kotlin/com/malinskiy/marathon/analytics/metrics/remote/RemoteDataSource.kt
+++ b/core/src/main/kotlin/com/malinskiy/marathon/analytics/metrics/remote/RemoteDataSource.kt
@@ -1,0 +1,15 @@
+package com.malinskiy.marathon.analytics.metrics.remote
+
+import java.time.Instant
+
+data class ExecutionTime(val testName: String,
+                         var percentile: Double)
+
+data class SuccessRate(val testName: String,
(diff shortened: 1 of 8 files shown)


Malinskiy:
@tagantroy can we add at least one test to check the expected behaviour which:
1. Cache if missing
2. Return from cache if present with the same params
3. Call again if missing in cache because of new params

core/src/main/kotlin/com/malinskiy/marathon/analytics/metrics/remote/RemoteDataSource.kt
```diff
@@ -0,0 +1,13 @@
+package com.malinskiy.marathon.analytics.metrics.remote
+
+import java.time.Instant
+
+data class ExecutionTime(val testName: String, var percentile: Double)
```
Vacxe: Will be great to split it to different files

tagantroy:
Hello @Malinskiy.
I added 2 tests per metric:

1. Return from cache if present with the same params
2. Call again if missing in cache because of new params

> Cache if missing 

Covered by "Return from cache if present with the same params"

Malinskiy (approved):

Malinskiy:
Fixes #254

PR #217, opened 2019-04-14
gradle plugin should mark output folder as gradle task output

### LINKED ISSUE #198: Do not allow running multiple marathon Gradle tasks in parallel
Currently, if you run `marathon` task in Gradle project with multiple modules it can run all marathon tasks in parallel which leads to random test failures.

To fix this we apply one output for all marathon tasks:
```
task.outputs.file(new File(rootProject.getBuildDir(), "fake-output-to-prevent-parallel-test-execution"))
```

Probably, we should do something similar in the gradle plugin by default.

Malinskiy:
to prevent multiple executions of the task at the same time

Fixes #198 

FYI @lukaville

### DIFF
--- a/marathon-gradle-plugin/src/main/kotlin/com/malinskiy/marathon/MarathonRunTask.kt
+++ b/marathon-gradle-plugin/src/main/kotlin/com/malinskiy/marathon/MarathonRunTask.kt
@@ -33,7 +33,7 @@ open class MarathonRunTask : DefaultTask(), VerificationTask {
     var ignoreFailure: Boolean = false
 
     @OutputDirectory
-    lateinit var output: File
+    var fakeLockingOutput = File(project.rootProject.buildDir, "fake-marathon-locking-output")
 
     @TaskAction
     fun runMarathon() {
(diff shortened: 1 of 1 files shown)


marathon-gradle-plugin/src/main/kotlin/com/malinskiy/marathon/MarathonRunTask.kt
```diff
@@ -31,14 +32,16 @@ open class MarathonRunTask : DefaultTask(), VerificationTask {
...
+    lateinit var output: File
 
     @TaskAction
     fun runMarathon() {
         val instrumentationApk = testVariant.extractTestApplication()
         val applicationApk = applicationVariant.extractApplication()
 
         val baseOutputDir = if (extensionConfig.baseOutputDir != null) File(extensionConfig.baseOutputDir) else File(project.buildDir, "reports/marathon")
-        val output = File(baseOutputDir, flavorName)
+        output = File(baseOutputDir, flavorName)
```
lukaville: but this directory will be different for each gradle module/flavour, so it will not fix the problem.
the problem is in running tests for multiple modules/flavours at the same time.

let's say we have 1 emulator and 2 gradle modules with tests. if we run `./gradlew marathon` it will try to run tests for both modules in parallel on the same emulator.

it will be better to have some kind of global emulator lock mechanism for all marathon run tasks in one gradle execution to distribute tests from multiple modules more evenly, but if each module has tests count >= connected emulators count then it's not a major problem

lukaville (approved):
please check that there are no warnings from Gradle
as far as I remember there was a warning like "output directory already exists" because Gradle may create this directory automatically

Issue #223, opened 2019-05-08
Add support to recognise multiple instances of emulator


witchlock:
When `AndroidDevice().serialNumber` returns `realSerialNumber`, it takes `serialProp` over `serialNumber` if `serialProp` exists. This causes issue that on my environment it always return the same `serialNumber` for different emulators.

`16:39:40.235 [AndroidDeviceProvider-BootWait-1 @coroutine#2] DEBUG AndroidDeviceProvider - Device EMULATOR28X0X23X0 connected channel.isFull = false. Healthy = true`
`16:39:40.235 [AndroidDeviceProvider-BootWait-3 @coroutine#4] DEBUG AndroidDeviceProvider - Device EMULATOR28X0X23X0 connected channel.isFull = false. Healthy = true`
`16:39:40.283 [AndroidDeviceProvider-BootWait-3 @coroutine#4] DEBUG AndroidDeviceProvider - Device EMULATOR28X0X23X0 booted!`
`16:39:40.283 [AndroidDeviceProvider-BootWait-1 @coroutine#2] DEBUG AndroidDeviceProvider - Device EMULATOR28X0X23X0 booted!`

Move `serialNumber.takeIf` up before `serialProp.takeIf` could resolve this issue.

tagantroy:
Hello @witchlock. Thanks for reporting this issue. I just created a bugfix. #224

Malinskiy:
Hey @witchlock

This behaviour is intentional. If you move the real serial (which is present all the time) you basically ignore all the other types of serial that we expect. The order was chose specifically by the level of uniqueness it provides:
1. Custom **marathon.serialno** which user is fully responsible for
2. "ro.boot.serialno" which is supposed to be unique for phones but it's not unique for Android emulators unless you start them (manually)[https://developer.android.com/studio/run/emulator-commandline]. Keep in mind that this property is propagated at boot time which means that snapshot of the emulator will have a cached value until you cold boot it.

```bash
emulator @Nexus_5X_API_23 -prop ro.serialno=123456
```
3. hostname is generally unique for android devices
4. ddms's serialno that is not even propagated properly from the boot for some stages of connection
5. random UUID. fails after a reconnect since it's not persistent

The problem of unique Android ID is over all the stack of development, so for this specific use case I'd suggest either setprop'ing our custom **marathon.serialno** property or booting the devices the official way with **-prop ro.serialno=123456**

[...]

PR #224, opened 2019-05-08
Fix realDeviceSerial order

tagantroy:
Fix for https://github.com/Malinskiy/marathon/issues/223

### DIFF
--- a/vendor/vendor-android/src/main/kotlin/com/malinskiy/marathon/android/AndroidDevice.kt
+++ b/vendor/vendor-android/src/main/kotlin/com/malinskiy/marathon/android/AndroidDevice.kt
@@ -74,9 +74,9 @@ class AndroidDevice(val ddmsDevice: IDevice) : Device, CoroutineScope {
         val serialNumber = ddmsDevice.serialNumber
 
         marathonSerialProp.takeIf { it.isNotEmpty() }
+                ?: serialNumber.takeIf { it.isNotEmpty() }
                 ?: serialProp.takeIf { it.isNotEmpty() }
                 ?: hostName.takeIf { it.isNotEmpty() }
-                ?: serialNumber.takeIf { it.isNotEmpty() }
                 ?: UUID.randomUUID().toString()
     }
 


Malinskiy:
This behaviour is intentional. If you move the real serial (which is present all the time) you basically ignore all the other types of serial that we expect. The order was chose specifically by the level of uniqueness it provides:
1. Custom **marathon.serialno** which user is fully responsible for
2. "ro.boot.serialno" which is supposed to be unique for phones but it's not unique for Android emulators unless you start them (manually)[https://developer.android.com/studio/run/emulator-commandline]. Keep in mind that this property is propagated at boot time which means that snapshot of the emulator will have a cached value until you cold boot it.

```bash
emulator @Nexus_5X_API_23 -prop ro.serialno=123456
```
3. hostname is generally unique for android devices
4. ddms's serialno that is not even propagated properly from the boot for some stages of connection
5. random UUID. fails after a reconnect since it's not persistent

The problem of unique Android ID is over all the stack of development, so for this specific use case I'd suggest either setprop'ing our custom **marathon.serialno** property or booting the devices the official way with **-prop ro.serialno=123456**

We can't merge this change because doing so will not allow people to fallback to hostname or ro.boot.serialno and will break the existing setups.

tagantroy:
@Malinskiy Hello, I think the best solution here - provide configuration flag for this behaviour and disable it by default

Malinskiy:
@tagantroy sure, let's do this via a separate pr

hborders:
How do you set a `marathon.serialno` in an emulator? When I try, I get the following warnings in my output:
```
WARNING: unexpected '-prop' value ('marathon.serialno=123456'), only 'qemu.*' properties are supported
```

and I get nothing when I try to get the property:
```
adb shell getprop marathon.serialno
```

This is how I create my AVD:
```
avdmanager --verbose create avd --name ui-testing --package 'system-images;android-27;google_apis_playstore;x86' --device "Nexus 5X" --abi 'google_apis_playstore/x86
```

This is how I start my emulator:
```
emulator -verbose -no-snapshot -wipe-data -accel on -no-boot-anim -no-audio -no-window -avd ui-testing -memory 4000 -netfast -partition-size 2048 -prop marathon.serialno=123456 &
```

I also tried setting `-prop ro.serialno=123456` with similar results.

Malinskiy:
@hborders one of the possible ways is using `setprop`, for example `adb shell setprop ro.serialno 123456` on a running instance of emulator.

hborders:
Thanks so much for your help!

I'm creating the emulator the same way as above, and using `setprop` fails:

```
$ adb shell setprop ro.serialno 123456
setprop: failed to set property 'ro.serialno' to '123456'
```

I'm running a headless emulator on ubuntu on an AWS `i3.metal`:
```
$ sdkmanager --list
Installed packages:=====================] 100% Computing updates...             
  Path                                               | Version      | Description                             | Location                                           
  -------                                            | -------      | -------                                 | -------                                            
  build-tools;29.0.2                                 | 29.0.2       | Android SDK Build-Tools 29.0.2          | build-tools/29.0.2/                                
  build-tools;29.0.3                                 | 29.0.3       | Android SDK Build-Tools 29.0.3          | build-tools/29.0.3/                                
  build-tools;30.0.2                                 | 30.0.2       | Android SDK Build-Tools 30.0.2          | build-tools/30.0.2/                                
  cmake;3.10.2.4988404                               | 3.10.2       | CMake 3.10.2.4988404                    | cmake/3.10.2.4988404/                              
  cmdline-tools;2.1                                  | 2.1          | Android SDK Command-line Tools          | cmdline-tools/2.1/                                 
  emulator                                           | 30.0.26      | Android Emulator                        | emulator/                                          
  ndk-bundle                                         | 21.3.6528147 | NDK                                     | ndk-bundle/                                        
  ndk;21.0.6113669                                   | 21.0.6113669 | NDK (Side by side) 21.0.6113669         | ndk/21.0.6113669/                                  
  patcher;v4                                         | 1            | SDK Patch Applier v4                    | patcher/v4/                                        
  platform-tools                                     | 30.0.4       | Android SDK Platform-Tools              | platform-tools/                                    
  platforms;android-29                               | 5            | Android SDK Platform 29                 | platforms/android-29/                              
  system-images;android-27;google_apis_playstore;x86 | 3            | Google Play Intel x86 Atom System Image | system-images/android-27/google_apis_playstore/x86/
```

Malinskiy:
@hborders I suspect you are not be executing this as root

https://github.com/agoda-com/adb-butler/blob/1171a5f459d84e958f9f3c9b195208b5ee89e929/bin/device_setup.sh#L47

hborders:
No, I am not. I didn't realize this needed root.

Since this is a play store image, I thought I couldn't root it.

hborders:
Since this is a play store image, I don't think I can root it easily.

PR #235, opened 2019-06-19
Fix JUnit report format

tagantroy:
According to [specification](https://www.ibm.com/support/knowledgecenter/en/SSUFAU_1.0.0/com.ibm.rsar.analysis.codereview.cobol.doc/topics/cac_useresults_junit.html) xml report should start with <testsuites> instead of <testsuite>


Malinskiy:
@tagantroy which CI integration did you test the output against?

tagantroy:
@Malinskiy I can check only with gitlab, but it doesn't provide integration for junit results parsing.

PR #237, opened 2019-06-21
Always fail execution when no devices are connected

### LINKED ISSUE #234: Fail gradle task when no devices found and ignore failures is true
Probably, when we set ignore failures option to `true` we should still fail the task in case of `no devices found` error. Currently, it has a successful result even if the test execution is failed due to no devices exception:

```
[16:42:58]: ▸ > Task :product:Badoo:libraries:BadooNative:marathonReleaseAndroidTest
[16:42:58]: ▸ Run instrumentation tests ...-release-androidTest.apk for app null
[16:42:58]: ▸ Ignore failures: true
[16:43:27]: ▸ com.malinskiy.marathon.exceptions.NoDevicesException: No devices found.
[16:43:28]: ▸ BUILD SUCCESSFUL in 21m 9s
```

I think the most suitable use case for this option is to have the ability to ignore only test results (but not compilation/other errors) and fail the build on CI level based on test reports. In such a way, we can use features like [test muting](https://confluence.jetbrains.com/pages/viewpage.action?pageId=43679832) in TeamCity. Therefore I think we should fail the build in any other cases other than a test failure or at least provide an option to configure runner in that way.

Malinskiy:
Closes #234

@lukaville

### DIFF
--- a/core/src/main/kotlin/com/malinskiy/marathon/Marathon.kt
+++ b/core/src/main/kotlin/com/malinskiy/marathon/Marathon.kt
@@ -88,7 +88,10 @@ class Marathon(val configuration: Configuration) {
             log.error(th.toString())
 
             when(th) {
-                is NoDevicesException -> false
+                is NoDevicesException -> {
+                    log.warn { "No devices found" }
+                    false
+                }
                 else -> configuration.ignoreFailures
             }
         }
--- a/core/src/main/kotlin/com/malinskiy/marathon/Marathon.kt
+++ b/core/src/main/kotlin/com/malinskiy/marathon/Marathon.kt
@@ -4,6 +4,7 @@ import com.google.gson.Gson
 import com.malinskiy.marathon.analytics.Analytics
 import com.malinskiy.marathon.analytics.AnalyticsFactory
 import com.malinskiy.marathon.device.DeviceProvider
+import com.malinskiy.marathon.exceptions.NoDevicesException
 import com.malinskiy.marathon.execution.Configuration
 import com.malinskiy.marathon.execution.Scheduler
 import com.malinskiy.marathon.execution.TestParser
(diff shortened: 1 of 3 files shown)


core/src/main/kotlin/com/malinskiy/marathon/Marathon.kt
```diff
@@ -85,7 +86,11 @@ class Marathon(val configuration: Configuration) {
             runAsync()
         } catch (th: Throwable) {
             log.error(th.toString())
-            false
+
+            when(th) {
+                is NoDevicesException -> false
+                else -> configuration.ignoreFailures
```
lukaville, line 95: do we have any other exceptions that may occur during normal test execution? I think we should fail the build in case of any test execution error that was not caused by the tests itself. also, looks like now we will throw an exception if we have failed tests and ignoreFailures is true. maybe we should return result class here something like `(testFailures: Boolean, executionErrors: Boolean)`?

then we can throw GradleException outside if:
* `executionErrors` is true
* `cnf.ignoreFailures` is false and `testFailures` is true

If we do this, we can also customize the error message. Because now for example if we don't have devices we still print "Tests failed" error and if we haven't run any tests we can point to an empty report

Malinskiy: It's quite hard to distinguish between test failures and execution error in practice. NoDevices is a valid use-case that we can implement right now and I'd rather evolve as we find more exceptions than try to guess what can go wrong at this point.

I've added the `warn` message `No devices found` in case the `NoDevicesException` happens that should distinguish between no tests and no devices.

Vacxe (approved):

core/src/main/kotlin/com/malinskiy/marathon/Marathon.kt
lukaville: OK, but what about the case when we have failed tests (and no exceptions during the execution) and ignoreFailures is true?
Looks like currently marathon Gradle task throws an exception in this case but it should not

PR #252, opened 2019-08-03
Refactor analytics

Malinskiy:
Added koin for DI, the Spek framework doesn't work with it out of the box


core/build.gradle.kts
```diff
@@ -56,6 +55,7 @@ dependencies {
     testCompile(TestLibraries.testContainers)
     testCompile(TestLibraries.testContainersInflux)
     testImplementation(TestLibraries.mockitoKotlin)
+    testImplementation(TestLibraries.koin)
```
tagantroy, line 56: Do we really need koin for tests?

core/build.gradle.kts
```diff
@@ -48,6 +46,7 @@ dependencies {
     implementation(Libraries.slf4jAPI)
     implementation(Libraries.logbackClassic)
     implementation(Libraries.influxDbClient)
+    compile(Libraries.koin)
```
tagantroy: replace compile with implementation

core/src/main/kotlin/com/malinskiy/marathon/analytics/TrackerFactory.kt
```diff
@@ -0,0 +1,72 @@
...
+import org.koin.core.KoinComponent
+import org.koin.core.inject
+import java.io.File
+
+internal class TrackerFactory(private val configuration: Configuration,
+                              private val fileManager: FileManager,
+                              private val gson: Gson,
+                              private val timer: Timer): KoinComponent {
+
+    val track: Track by inject()
```
tagantroy: Please use constructor injection instead of field injection

core/src/main/kotlin/com/malinskiy/marathon/analytics/TrackerFactory.kt
```diff
@@ -0,0 +1,72 @@
...
+        track + mappingTracker
+
+        return delegatingTrackerInternal
+    }
+
+    private fun createInfluxDbTracker(config: InfluxDbConfiguration): InfluxDbTracker? {
+        val db = try {
+            InfluxDbProvider(config).createDb()
+        } catch (e: Exception) {
+            null
```
tagantroy, line 56: What do you think about Result<T,Err> type instead of nullable return type?

core/src/main/kotlin/com/malinskiy/marathon/analytics/external/influx/InfluxDbTracker.kt
```diff
@@ -0,0 +1,39 @@
...
+import com.malinskiy.marathon.analytics.internal.sub.TrackerInternal
+import com.malinskiy.marathon.execution.TestStatus
+import com.malinskiy.marathon.test.toSafeTestName
+import org.influxdb.InfluxDB
+import org.influxdb.dto.Point
+import java.util.concurrent.TimeUnit
+
+internal class InfluxDbTracker(private val influxDb: InfluxDB) : TrackerInternal {
+    override fun track(event: Event) = when {
+        event is TestEvent &&
```
tagantroy: Should we wrap this type check in some generic interface?

core/src/main/kotlin/com/malinskiy/marathon/analytics/internal/pub/Track.kt
```diff
@@ -0,0 +1,52 @@
...
+import com.malinskiy.marathon.device.DevicePoolId
+import com.malinskiy.marathon.execution.TestResult
+import java.time.Instant
+import java.util.concurrent.atomic.AtomicReference
+
+class Track : Tracker {
+    private val delegates: AtomicReference<MutableList<Tracker>> = AtomicReference(mutableListOf())
+
+    operator fun plus(tracker: Tracker): Track {
+        delegates.getAndUpdate { list ->
```
tagantroy, line 14: I'm not sure about thread safety here

core/src/main/kotlin/com/malinskiy/marathon/execution/Scheduler.kt
```diff
@@ -33,7 +36,9 @@ class Scheduler(private val deviceProvider: DeviceProvider,
                 private val configuration: Configuration,
                 private val shard: TestShard,
                 private val progressReporter: ProgressReporter,
-                override val coroutineContext: CoroutineContext) : CoroutineScope {
+                override val coroutineContext: CoroutineContext) : CoroutineScope, KoinComponent {
+
+    private val track: Track by inject()
```
tagantroy: same here

core/src/main/kotlin/com/malinskiy/marathon/execution/queue/TestResultReporter.kt
```diff
@@ -1,21 +1,25 @@
...
+import org.koin.core.KoinComponent
+import org.koin.core.inject
 
 class TestResultReporter(private val poolId: DevicePoolId,
                          private val analytics: Analytics,
                          shard: TestShard,
-                         private val configuration: Configuration) {
+                         private val configuration: Configuration): KoinComponent {
+
+    private val track: Track by inject()
```
tagantroy: and here

core/src/main/kotlin/com/malinskiy/marathon/execution/strategy/SortingStrategy.kt
```diff
@@ -1,8 +1,8 @@
 package com.malinskiy.marathon.execution.strategy
 
-import com.malinskiy.marathon.analytics.metrics.MetricsProvider
+import com.malinskiy.marathon.analytics.external.MetricsProvider
 import com.malinskiy.marathon.test.Test
-import java.util.Comparator
+import java.util.*
```
tagantroy: why use a wildcard import instead of Comparator import?

core/src/test/kotlin/com/malinskiy/marathon/execution/queue/QueueActorSpek.kt
```diff
@@ -52,41 +58,32 @@ class QueueActorSpek : Spek({
...
-                    }
+            it("should report test as failed") {
+                declareMock<Track> {}
+                val track: Track by inject()
+                runBlocking {
+                    actor.send(QueueMessage.RequestBatch(TEST_DEVICE_INFO))
+                    poolChannel.receive()
+                    actor.send(QueueMessage.Completed(TEST_DEVICE_INFO, results))
+                    verify(track).test(any(), any(), captor.capture(), any())
+                    captor.firstValue.test shouldBe TEST_1
```
tagantroy, line 86: Do we really need to have this whitebox tests? Why not test it as a black box?

core/src/test/kotlin/com/malinskiy/marathon/execution/queue/TestResultReporterTest.kt
```diff
@@ -22,6 +26,7 @@ import org.jetbrains.spek.api.dsl.on
 import java.io.File
 
 object TestResultReporterSpec : Spek({
+    initKoin()
```
tagantroy: it is not needed with constructor injection

vendor/vendor-android/src/test/kotlin/com/malinskiy/marathon/android/spek/DeclareMock.kt
```diff
@@ -0,0 +1,27 @@
...
+import org.koin.core.qualifier.Qualifier
+import org.koin.test.mock.declareMockedDefinition
+import org.koin.test.mock.getDefinition
+
+/**
+ * Declare & Create a mock in Koin container for given type
+ *
+ * @author Arnaud Giuliani
+ */
+inline fun <reified T : Any> TestBody.declareMock(
```
tagantroy: Duplicate core/src/test/kotlin/com/malinskiy/marathon/spek/DeclareMock.kt

tagantroy (changes requested):

core/build.gradle.kts
Malinskiy: Yeah, we need now to inject analytics in every scenario tests by using `fun Spec.initKoin()`

core/build.gradle.kts
Malinskiy: this is intentional to propagate koin to other modules since injection takes place in CLI and gradle-plugin

core/src/main/kotlin/com/malinskiy/marathon/analytics/TrackerFactory.kt
Malinskiy: I've added logger error here. If you want to add functional style return types than I need to understand what to do about them later.

vendor/vendor-android/src/test/kotlin/com/malinskiy/marathon/android/spek/DeclareMock.kt
Malinskiy: Yeah. Do you want to share this a submodule?

core/src/test/kotlin/com/malinskiy/marathon/execution/queue/TestResultReporterTest.kt
Malinskiy: not sure how it would work then - this object is created by the test runner

core/src/test/kotlin/com/malinskiy/marathon/execution/queue/QueueActorSpek.kt
Malinskiy: I don't think having more tests is a bad thing. Removing tests is fine when they're not checking anything

core/src/main/kotlin/com/malinskiy/marathon/execution/strategy/SortingStrategy.kt
Malinskiy: no idea why IntelliJ chose this. fixed

core/src/main/kotlin/com/malinskiy/marathon/analytics/external/influx/InfluxDbTracker.kt
Malinskiy: you mean something like a InstanceTypeChecker?

core/src/main/kotlin/com/malinskiy/marathon/analytics/internal/pub/Track.kt
Malinskiy: the population happens in a single thread anyways, but even if not - what's the problem?

core/build.gradle.kts
Malinskiy: I've tried to remove it as much as possible from tests

core/src/main/kotlin/com/malinskiy/marathon/analytics/TrackerFactory.kt
Malinskiy: done

vendor/vendor-android/src/test/kotlin/com/malinskiy/marathon/android/spek/DeclareMock.kt
Malinskiy: Since I've removed koin from these tests I've deleted the android version, so there is only one left in core

core/build.gradle.kts
tagantroy: use `api` instead of `compile`, because compile is deprecated

core/src/main/kotlin/com/malinskiy/marathon/analytics/TrackerFactory.kt
tagantroy: In this case, if we ignore the error(only print it to logs) it is fine, but for some cases, it would be really useful to have Result type.

tagantroy (approved):

PR #256, opened 2019-08-23
Fix ignoreFailures behaviour in gradle plugin

### LINKED ISSUE #255: IgnoreFailures doesn't work
When I launch marathon by gradle plugin. I set ignoreFailures with true. But it doesn't work.

image

image

When any test failed, gradle task failed too

Malinskiy:
Fixes #255


marathon-gradle-plugin/src/main/kotlin/com/malinskiy/marathon/MarathonRunTask.kt
```diff
@@ -82,7 +82,8 @@ open class MarathonRunTask : DefaultTask(), VerificationTask {
 
         val success = Marathon(cnf).run()
 
-        if (!success) {
+        val shouldReportFailure = !cnf.ignoreFailures
```
tagantroy, line 85: 👍

tagantroy (approved):

Issue #142, opened 2018-12-26
Hanging tests are being returned to queue endlessly


Tieru:
To reproduce this issue you can set "testOutputTimeoutMillis" to very low value. Thus your tests won't be able to complete in time and begin rescheduling forever.

This happens because after "DevicePoolMessage.FromDevice.ReturnTestBatch" message all the tests are being rescheduled again and again without taking into account retry strategy (or similar) configurations.

Tieru:
This seems to be fixed (at least for us), although it's hard to say which changes solved this

Malinskiy:
This is fixed by #247, you can now set a quota and completely ignore these tests. By default runner will still continue to implement the "try until succeeds/fails" behaviour for backwards compatiblity

PR #262, opened 2019-08-28
Fix failing tests are being returned to queue endlessly #142

Atmosfearzx:
Issue #142 is not fixed. If test can't be finished (for example it's stuck), marathon is running endlessly. All in all, after rethrowing exceptions to TestBatchExecutionException we return tests to the queue. On other hand, we receive testRunEnded and proceed test as failed. Actually, there are a lot of solutions, but I just present one of them using parameter uncompletedTestRetryCount.

### DIFF
--- a/core/src/main/kotlin/com/malinskiy/marathon/execution/queue/QueueActor.kt
+++ b/core/src/main/kotlin/com/malinskiy/marathon/execution/queue/QueueActor.kt
@@ -81,6 +81,9 @@ class QueueActor(private val configuration: Configuration,
             handleFinishedTests(finished, device)
         }
         if (failed.isNotEmpty()) {
+            failed.forEach {
+                uncompletedTestsRetryCount[it.test] = (uncompletedTestsRetryCount[it.test] ?: 0) + 1
+            }
             handleFailedTests(failed, device)
         }
         if (uncompleted.isNotEmpty()) {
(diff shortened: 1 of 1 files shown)


Malinskiy:
Hey @Atmosfearzx. I've double checked the expected behaviour in https://github.com/Malinskiy/marathon/blob/ffd13bcc8f30f6707c302b1cb0108114efc55dda/core/src/test/kotlin/com/malinskiy/marathon/scenario/UncompletedScenarios.kt, there are some issues related to reporting uncompleted tests, but the test run finishes when you specify the uncompleted tests limitation. Can you please clarify what is your case and possible add a test which fails?
Thanks

Atmosfearzx:
Hello @Malinsky, I can provide the test only the next week, but there are certain steps to reproduce. Actually, I forgot to tell, that we use Android emualtors, so behaviour can be different.
1) You specify very low `testOutputTimeoutMillis` when some tests can't be finished in such short peroid of time. `uncompletedTestRetryQuota` doesn't affect it at all, but can be specified at low value to be sure.
2) Run marathon (tested on few android emulators)
3) Test runs and fails by timeout. In this case, we get log _"Test got stuck. You can increase the timeout in settings if it's too strict"_. The `TimeoutException` is catched in `AndroidDeviceTestRunner` and rethrowed as `TestBatchExceutionException`. 
4) After it `DeviceActor` catches this exception and calls `returnBatch`, test batch instantly returned to `queue` again (from this scenario you can see the test is gonna be added to queue every time it's failed by timeout and uncompletedTestRetryQuota can't rule this behaviour).

Atmosfearzx:
Hello, I see, you alredy fixed the issue the same way in https://github.com/Malinskiy/marathon/commit/9c3dad38d94edb44c1493c55f8847f10d9cbdbec Thank you very muchh. See you again.

Malinskiy:
no worries @Atmosfearzx! :) thanks for submitting

PR #265, opened 2019-08-30
Allow to disable HostKey verification for iOS

alex-apriamashvili:
#### CHANGELOG:
- [Add] `--rm` for every volume mapping in `docker run` command
- [Add] `disableHostKeyVerifier` configuration parameter to turn HostKey verification on demand.
- [Fix] an issue with loading default `known_hosts` even if the custom path is set.

### DIFF
--- a/cli/src/main/kotlin/com/malinskiy/marathon/cli/args/FileIOSConfiguration.kt
+++ b/cli/src/main/kotlin/com/malinskiy/marathon/cli/args/FileIOSConfiguration.kt
@@ -35,6 +35,7 @@ data class FileIOSConfiguration(
         @JsonProperty("keepAliveIntervalMillis") val keepAliveIntervalMillis: Long = 0L,
         @JsonProperty("deviceInitializationTimeoutMillis") val deviceInitializationTimeoutMillis: Long?,
         @JsonProperty("devices") val devices: File?,
+        @JsonProperty("disableHostKeyVerifier") val disableHostKeyVerifier: Boolean = false,
         val fileListProvider: FileListProvider = DerivedDataFileListProvider) : FileVendorConfiguration {
 
     fun toIOSConfiguration(marathonfileDir: File,
@@ -74,7 +75,8 @@ data class FileIOSConfiguration(
                     devicesFile = optionalDevices,
                     sourceRootsRegex = sourceRootsRegex,
                     sourceTargetName = sourceTargetName,
-                    binaryParserDockerImageName = binaryParserDockerImageName)
+                    binaryParserDockerImageName = binaryParserDockerImageName,
+                    disableHostKeyVerifier = disableHostKeyVerifier)
         } else {
             IOSConfiguration(
                     derivedDataDir = resolvedDerivedDataDir,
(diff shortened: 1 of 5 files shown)


vendor/vendor-ios/src/main/kotlin/com/malinskiy/marathon/ios/DockerTestParser.kt
```diff
@@ -24,20 +24,18 @@ class BinaryTestParser(private val binaryParserDockerImageName: String): DockerT
...
 
         val command =
-                listOf("docker", "run", "--rm") +
-                    pathMappings.map {
-                    listOf("-v", "${it.first}:${it.second}:ro") }
-                        .flatten() + binaryParserDockerImageName + pathMappings.map { it.second }
-
-        val builder = ProcessBuilder()
-                .command(command)
-                .directory(File(System.getProperty("user.dir")))
```
Malinskiy, line 37: why did we remove this directory as working folder?

cli/src/main/kotlin/com/malinskiy/marathon/cli/args/FileIOSConfiguration.kt
```diff
@@ -35,6 +35,7 @@ data class FileIOSConfiguration(
         @JsonProperty("keepAliveIntervalMillis") val keepAliveIntervalMillis: Long = 0L,
         @JsonProperty("deviceInitializationTimeoutMillis") val deviceInitializationTimeoutMillis: Long?,
         @JsonProperty("devices") val devices: File?,
+        @JsonProperty("disableHostKeyVerifier") val disableHostKeyVerifier: Boolean = false,
```
Malinskiy, line 38: For now this is an okay solution, but we need to think of a way to do the proper verification later

vendor/vendor-ios/src/main/kotlin/com/malinskiy/marathon/ios/cmd/remote/SshjCommandExecutor.kt
```diff
@@ -86,11 +89,13 @@ class SshjCommandExecutor(connectionId: String,
 
         try {
             ssh = SSHClient(config)
+            if (disableHostKeyVerifier) {
```
Malinskiy, line 93: :+1:
Can we emit an info log message for this?

Malinskiy (changes requested):

cli/src/main/kotlin/com/malinskiy/marathon/cli/args/FileIOSConfiguration.kt
alex-apriamashvili: the verification is on by default, this flag just allows to disable it for some cases. 
as for the verification on its own, it's done on `SSHj` side and I'm not sure if we can enhance it from within the marathon.

vendor/vendor-ios/src/main/kotlin/com/malinskiy/marathon/ios/DockerTestParser.kt
alex-apriamashvili: that's only about a docker process and it could be run from anywhere, so I see no reason in specifying this dir here.

vendor/vendor-ios/src/main/kotlin/com/malinskiy/marathon/ios/cmd/remote/SshjCommandExecutor.kt
alex-apriamashvili: will sure do

Malinskiy:
@alex-apriamashvili awesome work :+1:

PR #267, opened 2019-08-31
Uncompleted tests should be reported in any case

Malinskiy:


vendor/vendor-test/src/main/kotlin/com/malinskiy/marathon/test/StubDevice.kt
```diff
@@ -43,6 +45,10 @@ class StubDevice(
     ) {
         delay(testTimeMillis)
 
+        if (crashWithTestBatchException) {
+            throw TestBatchExecutionException("why not?")
```
tagantroy: The exception message is useless. Please add more information

Malinskiy: fixed

tagantroy (approved):

PR #281, opened 2019-10-24
remove CollectingShellOutputReceiver

tagantroy:
Use ddmlib CollectingOutputReceiver instead of CollectingShellOutputReceiver


Malinskiy:
@tagantroy Can I ask what's the original issue that we're fixing here?

tagantroy:
@Malinskiy I see no reason to use our own CollectingShellOutputReceiver if CollectingOutputReceiver already provided by ddmlib.

Malinskiy:
@tagantroy there is a difference of the expected string encoding though: ISO-8859-1 vs UTF-8

Issue #251, opened 2019-08-02
Logcat log from one test sometimes contains logs from several tests 


PStrelchenko:
Hi! Faced with strange issue with LogCat logs. 

Platform: Android

Assume we have a lot of tests with names T1 and T2, etc. In Allure's reports we see attached LogCat logs for each of this tests. But we noticed that logs of one test sometimes contains LogCat logs from another tests. 

It is not good - it makes understanding cause of errors really a mess. 

Maybe cause of this behavior lays in `com.malinskiy.marathon.android.executor.listeners.LogCatListener`

Malinskiy:
Hey @PStrelchenkoHH 

The logcat listener works in async fashion that is it is fired as soon as we know that test actually started.

What you describe can have multiple root causes:
- Logcat is buffer based so there might be a delay and you're seeing previous test logs
- Adb connection is laggy and even though start of the test already happened the test logs are still old
- Log reading thread might be reused on marathon side

To help you further could you please tell me more about one specific instance of this with the following info from one execution device:
- Timestamps of test executions
- Logcat logs with timestamps

at least a couple of batches of tests should be fine

Malinskiy:
After investigation the source of the error is the asynchronous nature of logcat request: whenever we call logcat it doesn't really filter the data for a specific timeframe.

I'm trying to address this with a separate project, will update this issue when there is a solution

Malinskiy:
This issue should be much less probable with the new vendor-adam in the develop branch.

The process of collecting the logcat log outlives the batch lifetime and is always in the background. The delay between sending command for collecting the log and actual collection is close to 0 in this case.

Malinskiy:
@PStrelchenko would you be able to test the latest version of the marathon with adam enabled and confirm if you experience the same problem?

Thanks

PStrelchenko:
@Malinskiy, ok, I'll try to give you feedback as soon as possible.
Thanks!

Malinskiy:
Bump

PStrelchenko:
Hi!
I've finally checked new Marathon with "adam" enabled, and as far as I know, there are no problems with logs anymore 👏 👏 👏 

Thanks!

PR #284, opened 2019-11-06
Feature -- Enable Kaspresso steps listener

PStrelchenko:
This PR adding new feature - enabling listener for test steps from [Kaspresso](https://github.com/KasperskyLab/Kaspresso) framework.

This works in the following way:

1. UI test created with TestCase from Kaspresso can hold a lot of steps with any description, e.g

```
@RunWith(AndroidJUnit4::class)
class OpenHomeScreenTest : TestCase() {

    @Rule
    @JvmField
    val runtimePermissionRule: GrantPermissionRule = GrantPermissionRule.grant(
            Manifest.permission.WRITE_EXTERNAL_STORAGE,
            Manifest.permission.READ_EXTERNAL_STORAGE
    )

    @Rule
    @JvmField
    val activityTestRule = ActivityTestRule(ApplicantRootActivity::class.java, true, false)

    @Test
    fun openHomeScreen() {
        before {
//            device.exploit.setOrientation(Orientation.Landscape)
            activityTestRule.launchActivity(null)

            waitSeconds(5)
        }.after {
//            device.exploit.setOrientation(Orientation.Portrait)
        }.run {
            step("Open Home Screen") {
                step("First sub step") {

                }
                step("Second sub step") {
                    step("Unexpected step") {

                    }
                }
            }
            step("Another step to show several steps") {

            }
        }
    }

}
```

Kaspresso itself [collects all steps](https://github.com/KasperskyLab/Kaspresso/pull/4), and when test finished, writes collected steps info JSON into log, something like this:

```
I/KASPRESSO: #AllureStepsInfoJson#: [{"attachments":[],"name":"My step 1","parameters":[],"stage":"finished","start":1568790287246,"status":"passed", "steps":[],"stop":1568790288184}]
```

This info then can be read with *LogCatListener* in Marathon. After reading, this JSON will be propagated into Marathon's *TestResult*. And after all, this JSON will be converted into Allure's steps and forwarded into Allure's report. 

===

This feature can be enabled with Marathon's configuration file, by setting special flag in Android's vendorConfiguration:

```
vendorConfiguration:
  type: "Android"
  ...
  enableKaspressoStepsListener: true
```

By default, this feature is *disabled*.

### DIFF
--- a/cli/src/main/kotlin/com/malinskiy/marathon/cli/args/FileAndroidConfiguration.kt
+++ b/cli/src/main/kotlin/com/malinskiy/marathon/cli/args/FileAndroidConfiguration.kt
@@ -2,6 +2,7 @@ package com.malinskiy.marathon.cli.args
 
 import com.fasterxml.jackson.annotation.JsonProperty
 import com.malinskiy.marathon.android.AndroidConfiguration
+import com.malinskiy.marathon.android.DEFAULT_ENABLE_KASPRESSO_STEPS_LISTENER
 import com.malinskiy.marathon.android.DEFAULT_INSTALL_OPTIONS
 import com.malinskiy.marathon.android.defaultInitTimeoutMillis
 import com.malinskiy.marathon.android.serial.SerialStrategy
@@ -20,7 +21,8 @@ data class FileAndroidConfiguration(
     @JsonProperty("adbInitTimeoutMillis") val adbInitTimeoutMillis: Int?,
     @JsonProperty("installOptions") val installOptions: String?,
     @JsonProperty("preferableRecorderType") val preferableRecorderType: DeviceFeature?,
-    @JsonProperty("serialStrategy") val serialStrategy: SerialStrategy = SerialStrategy.AUTOMATIC
+    @JsonProperty("serialStrategy") val serialStrategy: SerialStrategy = SerialStrategy.AUTOMATIC,
+    @JsonProperty("enableKaspressoStepsListener") val enableKaspressoStepsListener: Boolean?
 ) : FileVendorConfiguration {
 
     fun toAndroidConfiguration(environmentAndroidSdk: File?): AndroidConfiguration {
@@ -39,7 +41,8 @@ data class FileAndroidConfiguration(
             adbInitTimeoutMillis = adbInitTimeoutMillis ?: defaultInitTimeoutMillis,
             installOptions = installOptions ?: DEFAULT_INSTALL_OPTIONS,
             preferableRecorderType = preferableRecorderType,
-            serialStrategy = serialStrategy
+            serialStrategy = serialStrategy,
+            enableKaspressoStepsListener = enableKaspressoStepsListener ?: DEFAULT_ENABLE_KASPRESSO_STEPS_LISTENER
         )
     }
 }
--- a/cli/src/test/kotlin/com/malinskiy/marathon/cli/args/FileAndroidConfigurationSpek.kt
+++ b/cli/src/test/kotlin/com/malinskiy/marathon/cli/args/FileAndroidConfigurationSpek.kt
@@ -24,7 +24,8 @@ object FileAndroidConfigurationSpek : Spek(
                     null,
                     null,
                     null,
-                    SerialStrategy.AUTOMATIC
+                    SerialStrategy.AUTOMATIC,
+                    null
                 )
             }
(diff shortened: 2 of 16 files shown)


Vacxe:
@PStrelchenkoHH For me, that way looks like a plugin. In general, a test runner should abstract from 3rd party libraries. @Malinskiy isn't it?

Malinskiy:
Hey @PStrelchenkoHH 
Thanks for your PR. Having test steps in the report is a long-awaited feature and would be awesome to merge!

That said, I have several concerns with this change:
1. logcat as a transport layer is a bad decision because of the unreliable delivery. you were the one to report #251 as one of the cases where this fails. even though there is work going on regarding migration to [adam](https://github.com/Malinskiy/adam) there will always be the use case for ddmlib where this functionality will fail. That said there are still options to use on android that work reliably: [unknown test events are treated as metrics](https://android.googlesource.com/platform/tools/base/+/master/ddmlib/src/main/java/com/android/ddmlib/testrunner/InstrumentationResultParser.java#358) and can be gathered on the test runner side without any issues
2. I don't want to provide integration with a particular implementation of the test steps generator. Instead I'd prefer us to define a contract and a transport layer that we will use. It would be nice to have a designdoc for this and have a short discussion where everyone will talk about their view on this
3. There is no support for now for iOS in this hence it's even more important to define a contract that can be leveraged by other vendors. what should change is the transport layer but not the contract
4. Next time you do such a change I'd strongly encourage a short discussion before the implementation, that way we can speed up the merge, change and review process. Right now I'd like the above issues to be addressed before we merge this

Would you be able to draft a contract between a steps provider and runner as a consumer of this? Right now you emit the json, passed it as a string to allure code without any sanity checking. We should have an abstraction here to remove potential issues. I can also do this FYI

Our plan for this as I see right now:
1. Define contract for passing steps (independent of allure models)
2. Define transport how we get this for Android
3. This PR is changed to be an abstract consumer of the steps by implementing the contract
4. Kaspresso emits metrics by this standard

We can separate the publishing of steps into a module, that way it will be easier for you to not duplicate the code.

PStrelchenko:
Hey @Malinskiy, hey @Vacxe !
Thanks for your feedback on this feature! I'm encouraged that this feature will be helpful for Marathon's users =)

I've provided you our work solution and I agree, that current implementation is not so good, as it should be. But our team is ready to improve this solution for production-ready state =)

@Vacxe, yeap, this implementation is more like plugin or some temporary solution. And I believe that we can improve it. 

@Malinskiy, about your concerns:

1) I agree with you, LogCat as transport layer is not reliable solution (you were right with Issue # 251 - it was created after problems we encountered with propagating steps into Allure's reports through LogCat). Thanks for mentioning about ddmlib test metrics, it seems more suitable tool for our goal. 

2-3) Ok, it's sounds very reasonable, so let's try to define some abstract contract for producing and consuming steps information

4) Sorry, I'm not very familiar with contributing to open-source projects =) In future I will try to discuss solutions first and implement then =/

So, I will try to write design doc with proposed solution and move it into separate issue in this repository, so everyone could discuss it with us.

viclovsky:
Hello, @PStrelchenko , @Malinskiy, @Vacxe  
I use [allure-android ](https://github.com/allure-framework/allure-android) for my projects.  It is allure adaptor for android. It is designed as common listener (or android runner) and keeps allure models directly on device's sdcard.
It can be used with other runners, in which you are able to add custom listeners, and fraimeworks, like Kaspresso. 
Hope it will be helpfull.

Malinskiy:
For those interested in this: I've worked on https://github.com/Malinskiy/marathon/tree/feature/allure-steps several weeks ago. It works much better than the proposed integration with kaspresso.

Things that need work though:
1. The paclkage for `Epic` and other meta annotation lacks consistency. We have support for  io.qameta.allure, but the allure-android uses io.qameta.allure.android.annotations. The io.qameta.allure.android.annotations also is bundled with other android modules. I'd rather use them as a jar without any reliance and classpath from android.
2. Not possible to execute only one test with this custom runner. Needs support for @RunWith()
3. Due to asynchronous nature of disk io on Android the `/sdcard/allure-results` is empty after the execution of tests. This is the **biggest blocker for this approach**. This results in empty reports for a specific test. Marathon doesn't know anything about the expected files there and can't wait for them since the condition is unknown.

PR #285, opened 2019-11-08
refactor(vendor-android): create abstractions for new vendor-android implementation

Malinskiy:
As a start of integration with [adam](https://github.com/Malinskiy/adam) this PR separates ddmlib module into base module and ddmlib specific code


vendor/vendor-android/base/src/main/kotlin/com/malinskiy/marathon/android/RemoteFileManager.kt
```diff
@@ -0,0 +1,33 @@
+package com.malinskiy.marathon.android
+
+import com.malinskiy.marathon.test.Test
+
+class RemoteFileManager(private val device: AndroidDevice) {
```
Vacxe, line 5: Maybe will be better to extract interface here?

buildSrc/src/main/kotlin/Versions.kt
```diff
@@ -46,6 +46,8 @@ object Versions {
     val dokka = "0.9.17"
     val koin = "2.0.1"
     val jsonAssert = "1.5.0"
+    val ktor = "1.2.3"
```
tagantroy: Hey, I cannot find any usage of ktor. Do we need it?

vendor/vendor-android/base/src/main/kotlin/com/malinskiy/marathon/android/AndroidDevice.kt
```diff
@@ -0,0 +1,28 @@
+package com.malinskiy.marathon.android
+
+import com.android.sdklib.AndroidVersion
+import com.malinskiy.marathon.android.executor.listeners.line.LineListener
+import com.malinskiy.marathon.android.executor.listeners.video.ScreenRecorderOptions
+import com.malinskiy.marathon.device.Device
+import java.awt.image.BufferedImage
+import java.util.concurrent.TimeUnit
+
+interface AndroidDevice : Device {
```
tagantroy, line 10: Good job 👍

vendor/vendor-android/base/src/main/kotlin/com/malinskiy/marathon/android/AndroidTestParser.kt
```diff
@@ -28,10 +28,6 @@ class AndroidTestParser : TestParser {
             val className = packageAndClassName.substring(lastDotIndex + 1 until packageAndClassName.length)
 
             Test(packageName, className, methodName, annotations)
-        }.also {
-            if (it.isEmpty()) {
```
tagantroy, line 32: Why this code was removed?

vendor/vendor-android/base/src/main/kotlin/com/malinskiy/marathon/android/model/TestRunResultsAccumulator.kt
```diff
@@ -0,0 +1,196 @@
...
+     * Return the run failure error message, `null` if run did not fail.
+     */
+    var runFailureMessage: String? = null
+        private set
+
+    var aggregateMetrics = false
+
+    val completedTests: Set<TestIdentifier>
+        get() {
+            val completedTests = LinkedHashSet<TestIdentifier>()
```
tagantroy, line 40: Do we care about insertion order? If not, then why we need LinkedHashSet here?

buildSrc/src/main/kotlin/Versions.kt
Malinskiy: This is a requirement for adam. I can separate it in a different PR

vendor/vendor-android/ddmlib/src/main/kotlin/com/malinskiy/marathon/android/ddmlib/DdmlibAndroidDevice.kt
```diff
@@ -33,25 +53,127 @@ import kotlinx.coroutines.CompletableDeferred
...
+    }
+
+    private val nullOutputReceiver = NullOutputReceiver()
+
+    private val receiver = LogCatReceiverTask(ddmsDevice)
+
+    private var logcatThread = thread(name = "LogCatLogger-${ddmsDevice.serialNumber}") {
+        receiver.run()
+    }
+    private val ref = AtomicReference<MutableList<LogCatMessage>>(mutableListOf())
```
tagantroy: Why not to use immutable lists>

buildSrc/src/main/kotlin/Versions.kt
tagantroy: Wait, adam is separate library. If you define this dependency in adam i see no reason to define it here too

vendor/vendor-android/base/src/main/kotlin/com/malinskiy/marathon/android/AndroidTestParser.kt
Malinskiy: It was moved to Marathon.kt into the core

vendor/vendor-android/base/src/main/kotlin/com/malinskiy/marathon/android/model/TestRunResultsAccumulator.kt
Malinskiy: The original implementation of ddmlib [cares about the order](https://android.googlesource.com/platform/tools/base/+/tools_r22/ddmlib/src/main/java/com/android/ddmlib/testrunner/TestRunResult.java?autodive=0%2F%2F#38)

I think it's because some libs might do diff on the result.

vendor/vendor-android/base/src/main/kotlin/com/malinskiy/marathon/android/RemoteFileManager.kt
Malinskiy: I don't see this becoming standard between vendor modules for other platforms. For Android it's already using the AndroidDevice as an abstraction.
What would be the purpose of this?

buildSrc/src/main/kotlin/Versions.kt
Malinskiy: fixed

vendor/vendor-android/ddmlib/src/main/kotlin/com/malinskiy/marathon/android/ddmlib/DdmlibAndroidDevice.kt
Malinskiy: I removed this val completely, not used

vendor/vendor-android/base/src/main/kotlin/com/malinskiy/marathon/android/RemoteFileManager.kt
tagantroy: I'm agree with @Malinskiy. We already have abstraction here

PR #287, opened 2019-11-13
Add small optimization for ignored tests

tagantroy:
In case if all tests are ignored in batch and we have 0 tests to execute we don't need to cleanup data and start execution.


Malinskiy (approved):

Malinskiy:
Awesome improvement

PR #299, opened 2019-12-07
feat(gradle-plugin): add serialStrategy option

### LINKED ISSUE #298: Add support for setting serialStrategy using the Gradle Plugin
Heyho!

PR #231 introduced a new `serialStrategy` property to be able to explicitly choose how the serial of Android devices is derived. This is super useful in my CI environment where we create the required Android Emulator once and for each CI run start `n` read-only snapshot instances of this emulator to split up the tests.

However, I was wondering whether this `serialStrategy` can be set via the Gradle Plugin as well? I  couldn't find a matching property. Did I miss something?

Best regards,
Roland

Malinskiy:
Fixes #298

### DIFF
--- a/marathon-gradle-plugin/src/main/kotlin/com/malinskiy/marathon/MarathonExtension.kt
+++ b/marathon-gradle-plugin/src/main/kotlin/com/malinskiy/marathon/MarathonExtension.kt
@@ -1,5 +1,6 @@
 package com.malinskiy.marathon
 
+import com.malinskiy.marathon.android.serial.SerialStrategy
 import com.malinskiy.marathon.device.DeviceFeature
 import groovy.lang.Closure
 import org.gradle.api.Project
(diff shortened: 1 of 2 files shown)


CooperRS:
Heyho!

Thank you for opening the PR. Just gave it a try and it seems to work as expected. Only point I would have is that the API seems a bit weird. Could be improved with an import I guess 🙂.

```
marathon {
    codeCoverageEnabled = true
    autoGrantPermission = true
    serialStrategy = com.malinskiy.marathon.android.serial.SerialStrategy.DDMS
    batchingStrategy {
        fixedSize {
            size = 10
        }
    }
}
```

But in terms of MVP that absolutely fits my needs 👍

Thanks again!

Vacxe (approved):

Malinskiy:
Thanks for checking this!
I do prefer type-safety here rather than ease of configuration.
Will merge this

PR #301, opened 2019-12-13
feat(Android): Add support for code coverage

CooperRS:
## Description

This PR will add support for generating code coverage files while running tests on an Android device (for instance, an Android emulator) as well as downloading these files to the Marathon report directory after the tests have finished.

The main changes are:

1) Making the `AndroidTestRunner` read the `Configuration.isCodeCoverageEnabled` flag. If it is set, two instrumentation arguments are added to the command for executing tests (see [Android Documentation](https://developer.android.com/reference/androidx/test/runner/AndroidJUnitRunner)):
    * `coverage=true`, for enabling code coverage in general
    * `coverageFile=<path-to-coverage-file>`, for saving the coverage data at a known path
2) Adding a new `CodeCoverageListener` to the test run listeners for downloading the generated code coverage file to the Marathon report directory on the host computer. For this I created the new enum case `FileType.COVERAGE`, such that the coverage files would end up in the subfolder `coverage`.

However, when executing the above one needs to be careful about the `<path-to-coverage-file>`. The problem is that when tests are executed in batches each test batch is a new command which is executed on the Android device. Assuming the `<path-to-coverage-file>` is a static path, this would mean that each test batch would overwrite the coverage data from the previous test batch.

To prevent this, the `<path-to-coverage-file>` needs to include some dynamic parts making the path for each test batch unique. At first I tried to add an increasing number as a suffix to the path. However, I couldn't find such a number which would be available with in the `AndroidTestRunner`. My second attempt was to use the tests included in an instance of `TestBatch` to generate a unique id. However, this id was either not really unique or too long for a file name. Thus I settled with generating and assigning a UUID to each `TestBatch` instance upon creation. This is not guaranteed to be unique (as UUID are not guaranteed to be unique) but seemed to be unique enough (probability is super low).

## Screenshots

| Marathon Report |
|:---:|
| Bildschirmfoto 2019-12-13 um 12 38 35 |

## Open Questions

There are some known open questions that might need to be clarified before merging:

#### 1. Do you see the code coverage being merged into one file at some point?

Absolutely. In the project I'm currently working on we already have a Gradle task for doing this. This Gradle task was necessary anyways as the Android Test Orchestrator uses a similar approach to the one above and generates code coverage files for each test that is being run (with the above changes Marathon would do the same when the test batch size is set to 1). However, in order to generate XML/HTML coverage reports the developer has to merge these files as well. This can be done with a Gradle task like this:

```
task jacocoCoverageReport(type: JacocoReport) {
    group = "verification"
    description = "Converts JaCoCo coverage reports to XML/HTML reports."

    reports {
        xml.enabled = true
        html.enabled = true
    }

    sourceDirectories.from = files([sourceDir('app')])
    classDirectories.from = files([classDir('app')])
    executionData.from = fileTree(dir: rootDir, includes: ['**/*.exec', '**/*.ec'])
}
```

#### 2. What about potential clashes: one retry failed, another one succeeded for example?

I'm not perfectly sure, but I would assume that if the failed run covers the code paths A and B and the succeeded one covers the code paths A and C, the merged coverage files would mark the code paths A, B and C as covered.

As I did not touch existing code of Marathon, I would expect the HTML test report generated by Marathon to be unchanged compared to the current state 🙂.

### DIFF
--- a/vendor/vendor-android/base/src/main/kotlin/com/malinskiy/marathon/android/executor/listeners/CodeCoverageListener.kt
+++ b/vendor/vendor-android/base/src/main/kotlin/com/malinskiy/marathon/android/executor/listeners/CodeCoverageListener.kt
@@ -1,0 +1,30 @@
+package com.malinskiy.marathon.android.executor.listeners
+
+import com.malinskiy.marathon.android.AndroidDevice
+import com.malinskiy.marathon.android.model.TestIdentifier
+import com.malinskiy.marathon.device.DevicePoolId
+import com.malinskiy.marathon.device.toDeviceInfo
+import com.malinskiy.marathon.io.FileManager
+import com.malinskiy.marathon.io.FileType
+import com.malinskiy.marathon.log.MarathonLogging
+import com.malinskiy.marathon.test.TestBatch
+
+class CodeCoverageListener(
+    private val device: AndroidDevice,
+    private val devicePoolId: DevicePoolId,
+    private val fileManager: FileManager,
+    private val testBatch: TestBatch
+) : NoOpTestRunListener() {
+    private val logger = MarathonLogging.logger("CodeCoverageListener")
+
+    override suspend fun testRunEnded(elapsedTime: Long, runMetrics: Map<String, String>) {
+        val src = "${device.externalStorageMount}/coverage-${testBatch.id}.ec"
+        val dest = fileManager.createFile(FileType.COVERAGE, devicePoolId, device.toDeviceInfo(), testBatch.id).absolutePath
+
+        try {
+            device.pullFile(src, dest)
+        } catch (e: Exception) {
+            logger.error("Downloading coverage file $src failed!", e)
+        }
+    }
+}
--- a/vendor/vendor-android/ddmlib/src/main/kotlin/com/malinskiy/marathon/android/ddmlib/AndroidDeviceTestRunner.kt
+++ b/vendor/vendor-android/ddmlib/src/main/kotlin/com/malinskiy/marathon/android/ddmlib/AndroidDeviceTestRunner.kt
@@ -33,7 +33,7 @@ class AndroidDeviceTestRunner(private val device: DdmlibAndroidDevice) {
         val ignoredTests = rawTestBatch.tests.filter { test ->
             test.metaProperties.any { it.name == JUNIT_IGNORE_META_PROPERTY_NAME }
         }
-        val testBatch = TestBatch(rawTestBatch.tests - ignoredTests)
+        val testBatch = TestBatch(rawTestBatch.tests - ignoredTests, rawTestBatch.id)
 
         val androidConfiguration = configuration.vendorConfiguration as AndroidConfiguration
         val info = ApkParser().parseInstrumentationInfo(androidConfiguration.testApplicationOutput)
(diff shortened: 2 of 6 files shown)


CooperRS:
Any tips on how I can make the `Labeler` check green?

tagantroy:
@CooperRS I restarted job and it failed again, looks like something broken.
@Malinskiy can you check what's wrong with gitlab actions?

vendor/vendor-android/base/src/main/kotlin/com/malinskiy/marathon/android/executor/listeners/CodeCoverageListener.kt
```diff
@@ -0,0 +1,29 @@
...
+class CodeCoverageListener(
+    private val device: AndroidDevice,
+    private val devicePoolId: DevicePoolId,
+    private val fileManager: FileManager,
+    private val testBatch: TestBatch
+) : NoOpTestRunListener() {
+    private val logger = MarathonLogging.logger("CodeCoverageListener")
+
+    override fun testRunEnded(elapsedTime: Long, runMetrics: Map<String, String>) {
+        val src = "/sdcard/coverage-${testBatch.id}.ec"
```
tagantroy: is it possible to obtain this part using device.getExternalStorageMount()?

vendor/vendor-android/ddmlib/src/main/kotlin/com/malinskiy/marathon/android/ddmlib/AndroidDeviceTestRunner.kt
```diff
@@ -111,6 +111,10 @@ class AndroidDeviceTestRunner(private val device: DdmlibAndroidDevice) {
         androidConfiguration.instrumentationArgs.forEach { key, value ->
             runner.addInstrumentationArg(key, value)
         }
+        if (configuration.isCodeCoverageEnabled) {
+            runner.addInstrumentationArg("coverage", "true")
+            runner.addInstrumentationArg("coverageFile", "/sdcard/coverage-${testBatch.id}.ec")
```
tagantroy: same here

vendor/vendor-android/base/src/main/kotlin/com/malinskiy/marathon/android/executor/listeners/CodeCoverageListener.kt
CooperRS: Yep, that works as well. I'll amend 👍

Malinskiy:
The labeler fails because it's a fork apparently (https://github.com/actions/labeler/issues/12), apparently only disabling it will work

Malinskiy:
Hey @CooperRS, I've tried using your branch locally, and I suggest to add:
1. Some tips in logger if we failed to gather the code coverage in cases like missing WRITE_EXTERNAL_STORAGE permission or missing jacoco classes in classpath (i.e. when not using the `codeCoverageEnabled = true` option). Otherwise, it will be hard for users to understand what exactly is the problem and they'll post questions to marathon when they have incorrect setup
2. A code coverage setup in the docs (gh-pages) branch
3. Update the samples for app and library projects to include code coverage generation
3. Please double-check if everything works API 28+ (external storage API's have changed, want to make sure it works there)

Let me know if I can help with some of these.

Thanks

ZakTaccardi:
This would be super useful. Is this work still in progress @CooperRS? Thanks!

Malinskiy:
This feature implementation will be simplified after the merge of #444

Vacxe:
See: #466

PR #304, opened 2020-01-07
[Android] Fix cut off frames in GIF, Keep frames in the same target orientation

mikepenz:
* fix bug causing cut off images in the GIF if the screen rotates during a UI test case
  * detect target orientation based on the first frame
  * rotate follow up frames according to the initial frames detected orientation

### DIFF
--- a/vendor/vendor-android/base/src/main/kotlin/com/malinskiy/marathon/android/executor/listeners/screenshot/ScreenCapturer.kt
+++ b/vendor/vendor-android/base/src/main/kotlin/com/malinskiy/marathon/android/executor/listeners/screenshot/ScreenCapturer.kt
@@ -31,7 +31,7 @@ class ScreenCapturer(
     suspend fun start() = coroutineScope {
         val outputStream = FileImageOutputStream(fileManager.createFile(FileType.SCREENSHOT, poolId, device.toDeviceInfo(), test))
         val writer = GifSequenceWriter(outputStream, TYPE_INT_ARGB, DELAY, true)
-        var targetOrientation = PORTRAIT
+        var targetOrientation = UNDEFINED
         while (isActive) {
             val capturingTimeMillis = measureTimeMillis {
                 getScreenshot(targetOrientation)?.let {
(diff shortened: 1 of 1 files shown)


mikepenz:
@Malinskiy any suggestion what might be wrong with the labeler ?

BorisOsipov:
@mikepenz about labeler see https://github.com/Malinskiy/marathon/pull/301#issuecomment-566355128

mikepenz:
@BorisOsipov thank you very much

vendor/vendor-android/base/src/main/kotlin/com/malinskiy/marathon/android/executor/listeners/screenshot/ScreenCapturer.kt
```diff
@@ -31,9 +31,16 @@ class ScreenCapturer(
     suspend fun start() = coroutineScope {
         val outputStream = FileImageOutputStream(fileManager.createFile(FileType.SCREENSHOT, poolId, device.toDeviceInfo(), test))
         val writer = GifSequenceWriter(outputStream, TYPE_INT_ARGB, DELAY, true)
+        var targetOrientation = PORTRAIT
         while (isActive) {
             val capturingTimeMillis = measureTimeMillis {
-                getScreenshot()?.let { writer.writeToSequence(it) }
+                getScreenshot(targetOrientation)?.let {
+                    if (targetOrientation == UNDEFINED) {
```
Vacxe, line 38: Look like that this condition will be always ```false```.
Maybe you wanted to define ```var targetOrientation = PORTRAIT``` to ```UNDEFINED``` ?

mikepenz: Great catch. It should be actually UNDEFINED as long as it is unknown which orientation the images will have. based on the first screenshot we retrieve

Vacxe (approved):

Malinskiy:
For the future, we might consider recording just a square gif if we need to support cases with changing rotation better.

PR #308, opened 2020-01-08
[Android] Allow to configure to keep video on success

### LINKED ISSUE #306: [Android] Keep video captures also for success
In various scenarios it would be super helpful if we could keep also keep the test videos for success cases, especially when debugging test scenarios for devices which are only in a test rack and not accessible to the developer itself. 

I understand that in most cases this is not necessary and actually in various cases the available storage might be limited, as such keeping videos for all cases is not anticipated> 

As such a flag in the marathonconfig would be nice which would allow to retain the video if needed.

mikepenz:
- Introduce ScreenRecordingPolicy.kt 
  - integrate new global configuration for ScreenPolicy
  - add to test cases (1 default value, 1 interpreting value)
- add delete functionality for screenshot (GIF)
- delete GIF in case we don't want to keep it in SUCCESS state
  - only attach attachment if we kept the GIF

--------

- if not keeping ON_ANY or we had a failure
  - delete
- if we had ON_ANY
  - keep

Should fix: https://github.com/Malinskiy/marathon/issues/306

### DIFF
--- a/cli/src/main/kotlin/com/malinskiy/marathon/cli/args/FileConfiguration.kt
+++ b/cli/src/main/kotlin/com/malinskiy/marathon/cli/args/FileConfiguration.kt
@@ -2,6 +2,7 @@ package com.malinskiy.marathon.cli.args
 
 import com.malinskiy.marathon.execution.AnalyticsConfiguration
 import com.malinskiy.marathon.execution.FilteringConfiguration
+import com.malinskiy.marathon.execution.policy.ScreenRecordingPolicy
 import com.malinskiy.marathon.execution.strategy.BatchingStrategy
 import com.malinskiy.marathon.execution.strategy.FlakinessStrategy
 import com.malinskiy.marathon.execution.strategy.PoolingStrategy
@@ -37,7 +38,7 @@ data class FileConfiguration(
     var testOutputTimeoutMillis: Long?,
     var debug: Boolean?,
 
-    val screenRecordingPolicy: String?,
+    val screenRecordingPolicy: ScreenRecordingPolicy?,
 
     var vendorConfiguration: FileVendorConfiguration?,
 
--- a/cli/src/main/kotlin/com/malinskiy/marathon/cli/args/FileConfiguration.kt
+++ b/cli/src/main/kotlin/com/malinskiy/marathon/cli/args/FileConfiguration.kt
@@ -37,6 +37,8 @@ data class FileConfiguration(
     var testOutputTimeoutMillis: Long?,
     var debug: Boolean?,
 
+    val screenRecordingPolicy: String?,
+
     var vendorConfiguration: FileVendorConfiguration?,
 
     var analyticsTracking: Boolean?
--- a/core/src/main/kotlin/com/malinskiy/marathon/execution/policy/ScreenRecordingPolicy.kt
+++ b/core/src/main/kotlin/com/malinskiy/marathon/execution/policy/ScreenRecordingPolicy.kt
@@ -1,7 +1,5 @@
 package com.malinskiy.marathon.execution.policy
 
-import com.fasterxml.jackson.annotation.JsonCreator
-
 /**
  * Defines when screen recordings should be kept.
  */
@@ -14,17 +12,4 @@ enum class ScreenRecordingPolicy {
      * Keep screen recording in any case.
      */
     ON_ANY;
-
-    companion object {
-        @JvmStatic
-        @JsonCreator
-        fun fromString(key: String?): ScreenRecordingPolicy? {
-            return when (key) {
-                ON_ANY.name -> ON_ANY
-                ON_FAILURE.name -> ON_FAILURE
-                null -> null // in null case return nothing
-                else -> throw Exception("Wrong screen recording policy. Use one of ${values()}")
-            }
-        }
-    }
 }
--- a/core/src/main/kotlin/com/malinskiy/marathon/execution/policy/ScreenRecordingPolicy.kt
+++ b/core/src/main/kotlin/com/malinskiy/marathon/execution/policy/ScreenRecordingPolicy.kt
@@ -1,0 +1,28 @@
+package com.malinskiy.marathon.execution.policy
+
+import com.fasterxml.jackson.annotation.JsonCreator
+
+/**
+ * Defines when screen recordings should be kept.
+ */
+enum class ScreenRecordingPolicy {
+    /**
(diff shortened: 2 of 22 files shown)


Vacxe (approved):
In any case, this functionality can be used for some requirement

Malinskiy:
Hey @mikepenz! Thanks for the PR!
After some thought, I think this option should be for any screen capture, not only for video. As such, I think we should also save screenshots (`ScreenCapturerTestRunListener`).

A `ScreenRecordingPolicy` would be more suitable. Something like:

```kotlin
enum ScreenRecordingPolicy {
    ON_FAILURE, //Default
    ON_ANY
}
```

Then we use this in both `ScreenCapturerTestRunListener` and `ScreenRecorderTestRunListener` and independent of the recording type supported by the device we save the output for success if the user needs it.

What do you think?

vendor/vendor-android/ddmlib/src/main/kotlin/com/malinskiy/marathon/android/ddmlib/DdmlibAndroidDevice.kt
```diff
@@ -349,12 +350,12 @@ class DdmlibAndroidDevice(
     }
 
     private fun prepareRecorderListener(
-        feature: DeviceFeature, fileManager: FileManager, devicePoolId: DevicePoolId,
+        vendorConfiguration: VendorConfiguration, feature: DeviceFeature, fileManager: FileManager, devicePoolId: DevicePoolId,
         attachmentProviders: MutableList<AttachmentProvider>
     ): NoOpTestRunListener =
         when (feature) {
             DeviceFeature.VIDEO -> {
```
Malinskiy, line 363: This should also handle screenshots

Malinskiy (changes requested):

mikepenz:
@Malinskiy sounds good to me. I will do these changes during next week. (or perhaps tomorrow if I will get the time to do so) 

I haven't checked the iOS part of the code too much yet (as I only use it for testing android applications)
Shall it be a general vendor configuration or it is fine to keep it specific for android?

Malinskiy:
@mikepenz would be nice if we put it in general `Configuration` rather than vendor-specific. Thanks, mate. No worries about the timeline btw, whenever you have time.

mikepenz:
@Malinskiy updated the PR with the changes as you asked for.

core/src/main/kotlin/com/malinskiy/marathon/execution/policy/ScreenRecordingPolicy.kt
```diff
@@ -0,0 +1,28 @@
...
+    ON_ANY;
+
+    companion object {
+        @JvmStatic
+        @JsonCreator
+        fun fromString(key: String?): ScreenRecordingPolicy? {
+            return when (key) {
+                ON_ANY.name -> ON_ANY
+                else -> ON_FAILURE
+            }
```
Vacxe: @Malinskiy what is your suggestion here? 

Just for me fallback look strange here. 
ON_ANY.name -> ON_ANY
ON_FAILURE.name -> ON_FAILURE
else -> throw Exception("Wrong screen recording policy. Use one of ${ScreenRecordingPolicy.values()}")

vendor/vendor-android/base/src/main/kotlin/com/malinskiy/marathon/android/executor/listeners/video/ScreenRecorderTestRunListener.kt
```diff
@@ -69,7 +71,7 @@ class ScreenRecorderTestRunListener(
                 recorder?.join(awaitMillis)
             }
             logger.trace { "join ${join}ms" }
-            if (hasFailed) {
+            if (hasFailed || screenRecordingPolicy == ScreenRecordingPolicy.ON_ANY) {
```
Vacxe: ```hasFailed``` is subset of `ON_ANY`, will be better to swap the condition

core/src/main/kotlin/com/malinskiy/marathon/execution/policy/ScreenRecordingPolicy.kt
mikepenz: Ok let me change this

vendor/vendor-android/base/src/main/kotlin/com/malinskiy/marathon/android/executor/listeners/video/ScreenRecorderTestRunListener.kt
mikepenz: will update

Vacxe (approved):

cli/src/main/kotlin/com/malinskiy/marathon/cli/args/FileConfiguration.kt
```diff
@@ -37,6 +37,8 @@ data class FileConfiguration(
     var testOutputTimeoutMillis: Long?,
     var debug: Boolean?,
 
+    val screenRecordingPolicy: String?,
```
Malinskiy: Why is this a string? This is a POJO, the fact that we deserialize from string doesn't affect the data structure. Please keep the login of deserialization in a separate similar to `SortingStrategyDeserializer` :)

core/src/main/kotlin/com/malinskiy/marathon/execution/policy/ScreenRecordingPolicy.kt
Malinskiy: There is a code that does this already, it's `EnumDeserializer` in jackson. It also has options supported such as `READ_UNKNOWN_ENUM_VALUES_USING_DEFAULT_VALUE`. Can we please delegate this code to jackson?

Malinskiy (changes requested):

mikepenz:
@Malinskiy adjusted the code to use a deserializer instead of the manual string approach. 

I couldn't find a reference in the marathon code for the `READ_UNKNOWN_ENUM_VALUES_USING_DEFAULT_VALUE`  or using the `EnumDeserializer`. It also seemed as the compiler was complaining when trying to add the `JsonEnumDefaultValue` annotation to the enum

Malinskiy:
awesome job @mikepenz
and thank you for being patient with this long review

mikepenz:
@Malinskiy no worries. I highly appreciate accurate reviews and the high quality you and your other maintainers keep up with this project.

PR #318, opened 2020-01-13
Review atomic reference usage

### LINKED ISSUE #317: Review AtomicReference usage
I suggest to review how `AtomicReference` is used in these two places.
`com.malinskiy.marathon.analytics.internal.pub.Track#delegates` here I believe we have the https://en.wikipedia.org/wiki/ABA_problem and a data race between readers and writers.

`com.malinskiy.marathon.actor.StateMachine#stateRef` here notifications for state transitions can be dispatched out of order which may or may not be an issue.

Malinskiy:
Resolves #317

### DIFF
--- a/core/src/main/kotlin/com/malinskiy/marathon/actor/StateMachine.kt
+++ b/core/src/main/kotlin/com/malinskiy/marathon/actor/StateMachine.kt
@@ -29,27 +29,26 @@ package com.malinskiy.marathon.actor
 
 
 import java.util.*
-import java.util.concurrent.atomic.AtomicReference
 
 class StateMachine<STATE : Any, EVENT : Any, SIDE_EFFECT : Any> private constructor(
     private val graph: Graph<STATE, EVENT, SIDE_EFFECT>
 ) {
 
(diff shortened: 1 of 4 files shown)


core/src/main/kotlin/com/malinskiy/marathon/actor/StateMachine.kt
```diff
@@ -38,29 +38,33 @@ class StateMachine<STATE : Any, EVENT : Any, SIDE_EFFECT : Any> private construc
     private val stateRef = AtomicReference<STATE>(graph.initialState)
```
ash2k: This does not need to be an atomic reference - all accesses are protected by a synchronized section now and hence this can be just a normal field.

ash2k:
There are two additional things to consider:
1. Reading the `.state` field now can block while a transition is happening;
1. A transition is now in progress until all notifications have been dispatched.

Can any of that be an issue?

Vacxe (approved):

ash2k (approved):

Malinskiy:
@tagantroy would appreciate if you take a look at this since you might have some input for the state machine side-effects and possible deadlocks in this case

PR #319, opened 2020-01-14
Junit report refactor

rupinr:
This PR enables marathon to create a single Junit Report per pool/ per device comprising of all of the tests executed in them.
Previously marathon was creating one XML report per test case.

### DIFF
--- a/core/src/main/kotlin/com/malinskiy/marathon/io/FileManager.kt
+++ b/core/src/main/kotlin/com/malinskiy/marathon/io/FileManager.kt
@@ -29,6 +29,12 @@ class FileManager(private val output: File) {
         return File(resultsFolder, filename)
     }
 
+    fun createFile(fileType: FileType, pool: DevicePoolId, device: DeviceInfo, fileName: String): File {
+        val directory = createDirectory(fileType, pool, device)
+        val filename = createFilename(fileName, fileType)
+        return createFile(directory, filename)
+    }
+
     private fun createDirectory(fileType: FileType, pool: DevicePoolId, device: DeviceInfo): Path =
         createDirectories(getDirectory(fileType, pool, device))
 
@@ -47,6 +53,8 @@ class FileManager(private val output: File) {
     private fun createFile(directory: Path, filename: String): File = File(directory.toFile(), filename)
 
     private fun createFilename(test: Test, fileType: FileType): String = "${test.toTestName()}.${fileType.suffix}"
+
+    private fun createFilename(filename: String, fileType: FileType): String = "${filename}.${fileType.suffix}"
 
     private fun createFilename(device: DeviceInfo, fileType: FileType): String = "${device.serialNumber}.${fileType.suffix}"
 }
--- a/core/src/main/kotlin/com/malinskiy/marathon/io/FileManager.kt
+++ b/core/src/main/kotlin/com/malinskiy/marathon/io/FileManager.kt
@@ -29,12 +29,6 @@ class FileManager(private val output: File) {
         return File(resultsFolder, filename)
     }
 
-    fun createFile(fileType: FileType, pool: DevicePoolId, device: DeviceInfo, fileName: String): File {
-        val directory = createDirectory(fileType, pool, device)
-        val filename = createFilename(fileName, fileType)
-        return createFile(directory, filename)
-    }
-
     private fun createDirectory(fileType: FileType, pool: DevicePoolId, device: DeviceInfo): Path =
         createDirectories(getDirectory(fileType, pool, device))
 
@@ -53,8 +47,6 @@ class FileManager(private val output: File) {
     private fun createFile(directory: Path, filename: String): File = File(directory.toFile(), filename)
 
     private fun createFilename(test: Test, fileType: FileType): String = "${test.toTestName()}.${fileType.suffix}"
-
-    private fun createFilename(filename: String, fileType: FileType): String = "${filename}.${fileType.suffix}"
 
     private fun createFilename(device: DeviceInfo, fileType: FileType): String = "${device.serialNumber}.${fileType.suffix}"
 }
--- a/core/src/main/kotlin/com/malinskiy/marathon/report/junit/JUnitWriter.kt
+++ b/core/src/main/kotlin/com/malinskiy/marathon/report/junit/JUnitWriter.kt
@@ -1,16 +1,13 @@
 package com.malinskiy.marathon.report.junit
 
-import com.malinskiy.marathon.analytics.internal.sub.ExecutionReport
 import com.malinskiy.marathon.analytics.internal.sub.TestEvent
 import com.malinskiy.marathon.io.FileManager
 import com.malinskiy.marathon.io.FileType
 import com.malinskiy.marathon.report.junit.model.JUnitReport
 import com.malinskiy.marathon.report.junit.model.Pool
-import org.testng.util.Strings
 import java.io.FileOutputStream
 import javax.xml.stream.XMLOutputFactory
 import javax.xml.stream.XMLStreamWriter
-
 
 class JUnitWriter(private val fileManager: FileManager) {
 
--- a/core/src/main/kotlin/com/malinskiy/marathon/report/junit/JUnitWriter.kt
+++ b/core/src/main/kotlin/com/malinskiy/marathon/report/junit/JUnitWriter.kt
@@ -5,6 +5,7 @@ import com.malinskiy.marathon.device.DeviceInfo
 import com.malinskiy.marathon.device.DevicePoolId
 import com.malinskiy.marathon.io.FileManager
 import com.malinskiy.marathon.io.FileType
+import org.testng.util.Strings
 import java.io.FileOutputStream
 import javax.xml.stream.XMLOutputFactory
 import javax.xml.stream.XMLStreamWriter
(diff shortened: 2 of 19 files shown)


core/src/main/kotlin/com/malinskiy/marathon/report/junit/JUnitWriter.kt
```diff
@@ -1,75 +1,83 @@
...
+        }
+    }
 
-        val test = testResult.test
+    private fun prepareReport(executionReport: ExecutionReport) {
+        val reportGenerator = JunitReportGenerator(executionReport.testEvents)
+        reportGenerator.makeSuiteData()
+        val junitReports = reportGenerator.junitReports
+        junitReports.keys.forEach {
+            generateXml(xmlWriterMap[it]!!, junitReports[it]!!)
```
Vacxe: Unsafe null check

core/src/main/kotlin/com/malinskiy/marathon/report/junit/JUnitWriter.kt
```diff
@@ -1,75 +1,83 @@
...
+                junitReport.testCases.forEach {
+                    element("testcase") {
+                        attribute("classname", it.classname)
+                        attribute("name", it.name)
+                        attribute("time", it.time)
+                        if (it.skipped?.isNotEmpty()!!) {
+                            element("skipped", it.skipped)
                         }
-                        else -> {
+                        if (it.failure?.isNotEmpty()!!) {
```
Vacxe: Unsafe null check

core/src/main/kotlin/com/malinskiy/marathon/report/junit/JUnitWriter.kt
```diff
@@ -1,75 +1,83 @@
...
-                            element("failure") {
-                                writeCData(testResult.stacktrace ?: "")
-                            }
+
+                junitReport.testCases.forEach {
+                    element("testcase") {
+                        attribute("classname", it.classname)
+                        attribute("name", it.name)
+                        attribute("time", it.time)
+                        if (it.skipped?.isNotEmpty()!!) {
```
Vacxe: Unsafe null check

core/src/main/kotlin/com/malinskiy/marathon/report/junit/TestInfoGenerator.kt
```diff
@@ -0,0 +1,78 @@
...
+    fun makeSuiteData() {
+        val segregatedData = hashMapOf<DevicePool, List<TestEvent>>()
+        testEvents.devicePools().forEach { deviceData ->
+            segregatedData[deviceData] = testEvents.filter { it.device == deviceData.deviceInfo && it.poolId == deviceData.devicePool }
+        }
+        segregatedData.keys.forEach { deviceData ->
+            junitReports[deviceData] = JUnitReport(
+                segregatedData[deviceData]?.let { toTestSuite(it) }!!,
+                segregatedData[deviceData]?.let { toTestCase(it) }!!
+            )
```
Vacxe: Unsafe null checks

core/src/main/kotlin/com/malinskiy/marathon/report/junit/TestInfoGenerator.kt
```diff
@@ -0,0 +1,78 @@
...
+
+    private fun toTestSuite(testEvents: List<TestEvent>) =
+        TestSuiteData(
+            name = "common",
+            tests = testEvents.size,
+            failures = testEvents.filter { it.testResult.status == TestStatus.FAILURE }.size,
+            errors = 0,
+            skipped = testEvents.filter { it.testResult.status == TestStatus.IGNORED || it.testResult.status == TestStatus.ASSUMPTION_FAILURE }.size,
+            time = testEvents.map { it.testResult.durationMillis() }.sum().toJUnitSeconds(),
+            timeStamp = FORMATTER.format(testEvents.maxBy { it.testResult.endTime }?.testResult?.endTime!!)
```
Vacxe: Unsafe null check

core/src/main/kotlin/com/malinskiy/marathon/report/junit/JUnitWriter.kt
```diff
@@ -1,75 +1,83 @@
...
 import com.malinskiy.marathon.io.FileManager
 import com.malinskiy.marathon.io.FileType
+import org.testng.util.Strings
 import java.io.FileOutputStream
-import java.text.SimpleDateFormat
-import java.util.*
 import javax.xml.stream.XMLOutputFactory
 import javax.xml.stream.XMLStreamWriter
 
+private const val JUNIT_REPORT = "Marathon_Junit_Report"
```
Vacxe: Can be inner class constant

core/src/main/kotlin/com/malinskiy/marathon/report/junit/TestInfoGenerator.kt
```diff
@@ -0,0 +1,84 @@
...
+
+data class TestCaseData(
+    val classname: String,
+    val name: String,
+    val time: String,
+    val skipped: String,
+    val failure: String
+)
+
+fun List<TestEvent>.devicePools() = this.map { DevicePool(it.poolId, it.device) }.distinct()
```
Vacxe: I don't appreciate to use extension functions for geretic-specified types

core/src/main/kotlin/com/malinskiy/marathon/report/junit/TestInfoGenerator.kt
```diff
@@ -0,0 +1,84 @@
...
+)
+
+data class TestCaseData(
+    val classname: String,
+    val name: String,
+    val time: String,
+    val skipped: String,
+    val failure: String
+)
+
```
Vacxe: @Malinskiy @tagantroy 1 class - 1 file? Do u agree with that?

Vacxe (changes requested):

core/src/main/kotlin/com/malinskiy/marathon/report/junit/TestInfoGenerator.kt
rupinr: I too think that is better. I will change it.

Malinskiy:
Hey @rupinr . Thanks for the PR, awesome work 
I'd like to ask you to add some integration tests to this, something like adding sample events and then calling `JUnitReporter.generate` with an `ExecutionReport` with different events and then asserting that the resulting xml matches a fixture file.

The cases I'd like to see should include:
1. successful tests
2. failed tests
3. multiple executions of the same test that are changing from _failure_ to _success_ to failure in different order
4. other states like assumption failure or others from `TestStatus` enum

Thanks 🙏 , let me know if I can help clarify this

rupinr:
@Malinskiy Sure. Thanks for the review. I'll add the relevant tests and update the PR.

rupinr:
@Malinskiy  . I've updated the PR with relevant tests.

core/src/main/kotlin/com/malinskiy/marathon/report/junit/JUnitWriter.kt
```diff
@@ -1,72 +1,86 @@
...
-import com.malinskiy.marathon.device.DevicePoolId
-import com.malinskiy.marathon.execution.TestResult
-import com.malinskiy.marathon.execution.TestStatus
+import com.malinskiy.marathon.analytics.internal.sub.ExecutionReport
+import com.malinskiy.marathon.analytics.internal.sub.TestEvent
 import com.malinskiy.marathon.io.FileManager
 import com.malinskiy.marathon.io.FileType
+import com.malinskiy.marathon.report.junit.model.JUnitReport
+import com.malinskiy.marathon.report.junit.model.Pool
+import org.testng.util.Strings
```
Malinskiy: testng package in production code?

rupinr: Good catch :) Removed it.

core/src/main/kotlin/com/malinskiy/marathon/io/FileManager.kt
```diff
@@ -29,6 +29,12 @@ class FileManager(private val output: File) {
         return File(resultsFolder, filename)
     }
 
+    fun createFile(fileType: FileType, pool: DevicePoolId, device: DeviceInfo, fileName: String): File {
```
Malinskiy: We try to not have any fileName-dependent methods here. In this case the filename is also a constant `marathon_junit_report`. Would you please change this another method that doesn't have any file name?

Malinskiy (changes requested):
After the change with FileManager it's good to merge I think

core/src/main/kotlin/com/malinskiy/marathon/io/FileManager.kt
rupinr: Hi, I am bit confused here. But as I understand, I've removed the createFile with fileName and handled the file creation within junitWriter itself.

core/src/main/kotlin/com/malinskiy/marathon/report/junit/TestInfoGenerator.kt
```diff
@@ -0,0 +1,66 @@
...
+import java.util.*
+
+private val FORMATTER = SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss", Locale.US).apply {
+    timeZone = TimeZone.getTimeZone("UTC")
+}
+
+internal class JunitReportGenerator(private val testEvents: List<TestEvent>) {
+
+    @Suppress("MagicNumber")
+    private fun Long.toJUnitSeconds(): String = (this / 1000.0).toString()
```
Vacxe, line 21: TimeUnit.MILLISECONDS.toSeconds(this)

rupinr: But this will round of the seconds and I am not sure it is the expected behaviour.
 '35.757' would become '35'. Also, this is the existing behaviour.

rupinr:
> After the change with FileManager it's good to merge I think

Anything pending from my side?

Vacxe (approved):

Issue #320, opened 2020-01-14
Link timeline report


mikepenz:
One of the great reports marathon generates is the timeline report. 

Currently it is nowhere linked, and if you don't know the specific path it is hard to navigate to its location. 

Perhaps it will be helpful to link the timeline via the 
![image](https://user-images.githubusercontent.com/1476232/72364324-9f425280-36f6-11ea-8b6c-8682b61c3572.png)
duration field?

Alternative any other place would be as good. I think some visibility to it would be great, as it is a nice util to improve testing time

mikepenz:
Resolved with the merged PR

@Malinskiy please also to update the compiled html files :)

PR #321, opened 2020-01-14
Link timeline report via duration

mikepenz:
This PR is meant to improve upon: https://github.com/Malinskiy/marathon/issues/320

- link the timeline report via the duration field. 

To be honest, I am not 100% sure if this is the proper form of providing changes to the HTML section of this repo. 
I have installed the relevant node dependencies (actually ensured to have the old releases) 
And compiled the HTML. If it is preferred to only have the change in the raw js files, I can revert the result html.


html-report/src/components/HomePage.js
```diff
@@ -34,7 +34,7 @@ export default class HomePage extends Component {
                                     </div>
                                     <div className="card-info">
                                         <div className="text-sub-title-light">Duration</div>
-                                        <div className="card-info__content">{convertTime(pool.duration_millis)}</div>
```
Malinskiy: You're essentially adding a link to the timeline for all executed tests, but the duration is shown only for one pool of devices. Also discovering this would be problematic. Would it be ok if we just add the link to the top of the page near POOLS? I just don't think the context is correct in this case.

Malinskiy (changes requested):
Does it make sense to include the min.js change here? I agree we need to update it, but we don't have a good process for this now. I can't follow the diff for one-line file change of this size :(

html-report/src/components/HomePage.js
mikepenz: Sure I have no strong opinion on the location. I can change this around

mikepenz:
@Malinskiy ok thank you I'll remove the compiled html in this case and will only include the normal html change in this PR

mikepenz:
@Malinskiy what would you think about: 

image

Malinskiy:
@mikepenz even better! 🏆 this looks awesome

mikepenz:
Ok the PR is updated with this :) 
It only contains the single html change now. no longer the compiled result. 

I tested it with the current Chrome Beta.

Malinskiy (approved):

PR #325, opened 2020-01-20
fix(cli): exit 1 only if configuration allows this

Malinskiy:


mikepenz (approved):

cli/src/main/kotlin/com/malinskiy/marathon/cli/ApplicationView.kt
```diff
@@ -50,9 +50,11 @@ fun main(args: Array<String>): Unit = mainBody(
         UsageAnalytics.USAGE_TRACKER.trackEvent(Event(TrackActionType.RunType, "cli"))
         val success = marathon.run()
         bugsnagExceptionsReporter.end()
-        if (!success) {
+
+        val shouldReportFailure = !configuration.ignoreFailures
+        if (!success && shouldReportFailure) {
```
lukaville, line 55: but in this case, the build will not fail if the run was not successful because of other reasons (not related to failed tests). I think the desired behavior for most users is:

* when `ignoreFailures` = true, then do not fail on tests failures, but fail on other errors (such as no devices exception)
* when `ignoreFailures` = false, then fail in both cases (test failures and other errors)

then you can configure CI to fail on non-zero exit code in case of device problems (no devices exception) or marathon crashes but don't fail if there are failed tests (so it is possible to "mute" the test by using CI capabilities)

I think we should modify `Marathon.kt` to return false in case of not test-related errors and when ignoreFailures is enabled. Like this:

```
fun run() = runBlocking {
        try {
            val isSuccess = runAsync()
            when {
                configuration.ignoreFailures -> true
                else -> isSuccess
            }
        } catch (th: Throwable) {
            log.error(th.toString())

            when (th) {
                is NoDevicesException -> {
                    log.warn { "No devices found" }
                    false
                }
                else -> false
            }
        }
}
```

BorisOsipov: @lukaville comment seems reasonable. 

Also it looks like to have different exit codes for different cases e.g. failed tests\Marathon errors\no devices exceptions will be good - but it is out of scope of this issue for sure.

Malinskiy: I would be happy to introduce ignoreNoDevices and other types of ignoring exceptions, but this change only addresses the issue of inconsistency between CLI and Gradle plugin.

We can add the flag to specifically ignore test failures and then map exceptions to exit codes, however, I don't have a grasp on how to map this to Gradle plugin logic. I've created https://github.com/Malinskiy/marathon/issues/327 to address the issue that you brought up.

Issue #331, opened 2020-02-03
Build does not fail though count sharding and Strict mode is true


kristfoograb:
### Problem:
When using Count sharding and Strict mode true to run test for x number of times, Marathon does not throw SystemExitException though the test fail at least once.

The problem arise because PoolProgressTracker does not allow transition to Fail state if the test has passed previously. I believe the code was designed in this manner because test run should end with final state as pass. This is true prior to the introduction of StrictMode configuration https://github.com/Malinskiy/marathon/pull/236.

PS: This seems to be a similar issue raised in https://github.com/Malinskiy/marathon/issues/196 but the enhancement was only introduced to TestResultReporter.

### Proposed Change:
I believe changes need to be implemented in PoolProgressTracker.

```kotlin
        state<ProgressTestState.Passed> {
            on<ProgressEvent.Failed> {
                if (configuration.strictMode) {
                    transitionTo(ProgressTestState.Failed)
                }else{
                    dontTransition()
                }
            }
            on<ProgressEvent.Ignored> {
                dontTransition()
            }
        }
        state<ProgressTestState.Failed> {
            on<ProgressEvent.Passed> {
                if (configuration.strictMode) {
                    dontTransition()
                }else {
                    transitionTo(ProgressTestState.Passed)
                }
            }

        }
```

Malinskiy:
Have you been able to confirm that the merged PR version is working for your use-case @kristfoograb?

If yes, please close the issue. Thanks 👍

kristfoograb:
Thanks for merging.

PR #332, opened 2020-02-11
Build does not fail though count sharding and Strict mode is true

kristfoograb:
Pass configuration into PoolProgressTracker so that when strict mode is true, transition from pass to fail can happen. This is raised for issue https://github.com/Malinskiy/marathon/issues/331

### DIFF
--- a/core/src/main/kotlin/com/malinskiy/marathon/execution/progress/ProgressReporter.kt
+++ b/core/src/main/kotlin/com/malinskiy/marathon/execution/progress/ProgressReporter.kt
@@ -11,9 +11,8 @@ import kotlin.math.roundToInt
 
 const val HUNDRED_PERCENT_IN_FLOAT: Float = 100.0f
 
-class ProgressReporter {
+class ProgressReporter(private val configuration: Configuration) {
     private val reporters = ConcurrentHashMap<DevicePoolId, PoolProgressTracker>()
-    private lateinit var configuration: Configuration
 
     private inline fun <T> execute(poolId: DevicePoolId, f: (PoolProgressTracker) -> T): T {
         val reporter = reporters[poolId] ?: PoolProgressTracker(configuration)
@@ -51,10 +50,6 @@ class ProgressReporter {
         return reporters.isNotEmpty() && reporters.values.all {
             it.aggregateResult()
         }
-    }
-
-    fun init(configuration: Configuration){
-        this.configuration = configuration
     }
 
     fun totalTests(poolId: DevicePoolId, size: Int) {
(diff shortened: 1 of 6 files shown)


core/src/main/kotlin/com/malinskiy/marathon/execution/progress/ProgressReporter.kt
```diff
@@ -12,9 +13,10 @@ const val HUNDRED_PERCENT_IN_FLOAT: Float = 100.0f
 
 class ProgressReporter {
     private val reporters = ConcurrentHashMap<DevicePoolId, PoolProgressTracker>()
+    private lateinit var configuration: Configuration
```
Vacxe: I don't appreciate to use `lateinit` without any annotation processing stuff

core/src/main/kotlin/com/malinskiy/marathon/execution/progress/ProgressReporter.kt
```diff
@@ -51,6 +53,10 @@ class ProgressReporter {
         }
     }
 
+    fun init(configuration: Configuration){
```
Vacxe: `init` actually is a bad tone. This function looks like a setter for configuration.

core/src/main/kotlin/com/malinskiy/marathon/execution/queue/QueueActor.kt
```diff
@@ -50,6 +50,7 @@ class QueueActor(
 
     init {
         queue.addAll(testShard.tests + testShard.flakyTests)
+        progressReporter.init(configuration)
         progressReporter.totalTests(poolId, queue.size)
```
Vacxe: Maybe it make sense to add `configuration` as parameter to `totalTests` func

Vacxe (changes requested):

core/src/main/kotlin/com/malinskiy/marathon/execution/progress/ProgressReporter.kt
kristfoograb: I have removed this line and pass variable through constructor

core/src/main/kotlin/com/malinskiy/marathon/execution/progress/ProgressReporter.kt
kristfoograb: Removed.

core/src/main/kotlin/com/malinskiy/marathon/execution/queue/QueueActor.kt
kristfoograb: Removed as well.

Vacxe (approved):

Malinskiy (approved):

PR #335, opened 2020-02-29
Make reports independent of attachment naming

TrubnikovDmitriy:
I want to make reports independent of attachment names.
This add the ability to change the name of the attachment without breaking the reports.

I found two weak points:

1. **HtmlSummaryReporter**. It has already been refactored more than once, so I suggest a different solution. Instead of a hardcore path to the files and checking for their existence, just _take the attachments_.
2. **ScreenCapturerTestRunListener**. This only works because the same hardoced names are passed to the constructor of `File`. It is much more reliable to _pass a File_ to the `ScreenCapturer` where need to write a GIF.


TrubnikovDmitriy:
@Vacxe, @Malinskiy, hello!
What do you think about these changes?

PR #377, opened 2020-06-10
Update ddmlib to 27.0.0 to enable ANDROID_ADB_USER_MANAGED_MODE support and AGP 4.0

TimoPtr:
I'm running into an environment where the lifecycle of ADB need to be control outside the build to do that we need to update the DDMLIB to 27.0.0 that introduce a flag called sUserManagedAdbMode it's set through the env variable `export ANDROID_ADB_USER_MANAGED_MODE=true`

It will prevent the ADB server to be killed

### DIFF
--- a/buildSrc/src/main/kotlin/Versions.kt
+++ b/buildSrc/src/main/kotlin/Versions.kt
@@ -4,7 +4,7 @@ object Versions {
     val kotlin = "1.3.61"
     val coroutines = "1.3.1"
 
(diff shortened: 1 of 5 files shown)


TimoPtr:
Using this library force us to use AGP 4.0 because of StringHelper migration to Kotlin in common library

Vacxe (approved):
LGTM

tagantroy (approved):

PR #395, opened 2020-07-14
Change order for SerialStrategy.AUTOMATIC

tagantroy:
By default, marathonSerialProp is empty.
SerialProp, in my case, looks the same for all instances(EMULATOR29X2X1X0).
ADB serial number is the best option for default behavior


Malinskiy:
@tagantroy can you please check/update the docs to describe the default behaviour and the edge cases associated with this?

Malinskiy (approved):

tagantroy:
I created #397 with updates for documentation

Issue #400, opened 2020-09-21
Graphite support


bartek-wesolowski:
Marathon supports gathering test statistics using InfluxDB. You can then visualize them with Graphite.
It would be great to be able to use Graphite instead of InfluxDB as the underlying database. In our company we use Graphite for other statistics and we could integrate it with Marathon easily without the need of having to host another tool.

bartek-wesolowski:
I'm working on it. It should be done in a couple of days.

bartek-wesolowski:
I added Graphite support, but I don't seem to have the necessary repository permissions to push the changes. Here's the patch file. Please, give me the necessary permissions or apply it yourself.

[graphite.patch.zip](https://github.com/Malinskiy/marathon/files/5347515/graphite.patch.zip)

tagantroy:
Thanks for your contribution. 
#406 and #408 merged 
Please reopen if there anything else to do.

PR #406, opened 2020-10-08
Add Graphite support as an alternative to InfluxDB

bartek-wesolowski:


bartek-wesolowski:
Linked issue: https://github.com/Malinskiy/marathon/issues/400

cli/src/main/kotlin/com/malinskiy/marathon/cli/config/deserialize/GraphiteConfigurationDeserializer.kt
```diff
@@ -0,0 +1,29 @@
...
+        val portString = node?.get("port")?.asText()
+        val prefix = node?.get("prefix")?.asText()
+
+        val port = portString?.toIntOrNull()
+
+        if (host == null) throw ConfigurationException("GraphiteConfigurationDeserializer: host should be specified")
+        if (portString != null && port == null) throw ConfigurationException("GraphiteConfigurationDeserializer: port should be a number, e.g. 2003")
+        if (prefix?.isEmpty() == true) throw ConfigurationException("GraphiteConfigurationDeserializer: prefix cannot be empty")
+
+        return AnalyticsConfiguration.GraphiteConfiguration(host, port ?: 2003, prefix)
```
Malinskiy: Deserialiser is not responsible for handling default values. Please make the port optional (as in nullable) and handle the defaults outside, e.g. in the data class definition via constructor default argument value.

cli/src/main/kotlin/com/malinskiy/marathon/cli/config/deserialize/GraphiteConfigurationDeserializer.kt
```diff
@@ -0,0 +1,29 @@
...
+    override fun deserialize(p: JsonParser?, ctxt: DeserializationContext?): AnalyticsConfiguration.GraphiteConfiguration {
+        val node: JsonNode? = p?.codec?.readTree(p)
+
+        val host = node?.get("host")?.asText()
+        val portString = node?.get("port")?.asText()
+        val prefix = node?.get("prefix")?.asText()
+
+        val port = portString?.toIntOrNull()
+
+        if (host == null) throw ConfigurationException("GraphiteConfigurationDeserializer: host should be specified")
```
Malinskiy, line 23: The only reason AFAIU for this deserialiser to exist is custom error handling. Correct me if I'm wrong. Thanks.

core/src/main/kotlin/com/malinskiy/marathon/analytics/TrackerFactory.kt
```diff
@@ -37,10 +40,11 @@ internal class TrackerFactory(
     fun create(): TrackerInternal {
         val defaultTrackers = mutableListOf<TrackerInternal>(createExecutionReportGenerator())
 
-        if (configuration.analyticsConfiguration is InfluxDbConfiguration) {
-            val config = configuration.analyticsConfiguration
-            val influxDbTracker = createInfluxDbTracker(config)
-            influxDbTracker?.let { defaultTrackers.add(it) }
+        val analyticsConfiguration = configuration.analyticsConfiguration
```
Malinskiy, line 43: Would be easier to read if it was done using when, e.g.:

```kotlin
        val analyticsConfiguration = configuration.analyticsConfiguration
        val analyticsTracker = when (analyticsConfiguration) {
            is InfluxDbConfiguration -> createInfluxDbTracker(analyticsConfiguration)
            is GraphiteConfiguration -> createGraphiteTracker(analyticsConfiguration)
            else -> null
        }
        if (analyticsTracker != null) {
            defaultTrackers.add(analyticsTracker)
        }
```

core/src/main/kotlin/com/malinskiy/marathon/analytics/external/MetricsProviderFactory.kt
```diff
@@ -1,17 +1,34 @@
...
     private val configuration = configuration.analyticsConfiguration
 
     fun create(): MetricsProvider {
-        return if (configuration is AnalyticsConfiguration.InfluxDbConfiguration) {
-            InfluxMetricsProvider.createWithFallback(configuration, NoOpMetricsProvider())
-        } else {
-            NoOpMetricsProvider()
+        return when (configuration) {
+            is InfluxDbConfiguration -> createInfluxDBMetricsProvider(configuration, NoOpMetricsProvider())
+            is GraphiteConfiguration -> createGraphiteMetricsProvider(configuration)
```
Malinskiy, line 17: Even if the configuration is correct, there will be a case of network failure. The NoOpMetricsProvider fallback is there for a reason: when the graphite server will fail - we will fallback. The Influx implementation just does `try/catch` to verify everything is working. You should implement the same idea for graphite here.

core/src/main/kotlin/com/malinskiy/marathon/analytics/external/graphite/GraphiteUtils.kt
```diff
@@ -0,0 +1,9 @@
+package com.malinskiy.marathon.analytics.external.graphite
+
+internal fun String.withPrefix(prefix: String?): String {
```
Malinskiy: We have a convention to put every extension of default classes into `extension` package. Could you please move this to `com.malinskiy.marathon.extension`?

core/src/main/kotlin/com/malinskiy/marathon/analytics/external/graphite/BasicGraphiteClient.kt
```diff
@@ -0,0 +1,70 @@
...
+import java.io.IOException
+import java.io.PrintWriter
+import java.net.Socket
+import java.net.UnknownHostException
+import java.time.Instant
+
+data class GraphiteMetric(
+    val metricName: String,
+    val metricValue: String,
+    val timestamp: Long = Instant.now().toEpochMilli()
```
Malinskiy, line 20: All 3 instantiations of GraphiteMetric provide the timestamp value. Do we really need a default here?

core/src/main/kotlin/com/malinskiy/marathon/analytics/external/graphite/GraphiteTracker.kt
```diff
@@ -0,0 +1,51 @@
...
+import com.malinskiy.marathon.device.DeviceInfo
+import com.malinskiy.marathon.execution.TestStatus
+import com.malinskiy.marathon.test.Test
+import com.malinskiy.marathon.test.toSafeTestName
+
+class GraphiteTracker(
+    private val graphite: GraphiteClient
+) : TrackerInternalAdapter() {
+
+    override fun trackTest(event: TestEvent) {
```
Malinskiy, line 14: Graphite supports [tags](https://graphite.readthedocs.io/en/latest/tags.html)

We should use them to allow constructing dashboards and querying data in a simple manner. See the InfluxDbTracker to get the idea of the tags that you need to pass.

core/src/main/kotlin/com/malinskiy/marathon/analytics/metrics/remote/graphite/GraphiteDataSource.kt
```diff
@@ -0,0 +1,81 @@
...
+        val scanner = Scanner(connection.inputStream, StandardCharsets.UTF_8.name())
+        val lines = mutableListOf<String>()
+        while (scanner.hasNextLine()) {
+            lines.add(scanner.nextLine())
+        }
+        return lines
+    }
+
+    override fun close() {
+        // nop
```
Malinskiy, line 51: The current implementation of influx is addressing the fact that we query a lot of data from influx: it reuses the connection. This implementation of graphite will slow down tests. Is there any way we can reuse the connect and skip the TCP handshake for potentially thousands of requests?

core/src/main/kotlin/com/malinskiy/marathon/analytics/metrics/remote/graphite/GraphiteDataSource.kt
```diff
@@ -0,0 +1,81 @@
...
+            val duration = line.substringAfter('|').toDouble()
+            ExecutionTime(testName, duration)
+        }
+    }
+
+    private fun query(target: String, from: Instant): List<String> {
+        val encodedTarget = URLEncoder.encode(target, StandardCharsets.UTF_8.name())
+        val formattedFrom = fromFormatter.format(from)
+        val url = URL("http://${host}/render?target=$encodedTarget&format=raw&from=$formattedFrom")
+        val connection = url.openConnection() as HttpURLConnection
```
Malinskiy: We use okhttp for http connections, I suggest you rewrite this so as to not introduce another HTTP stack to the project.

cli/src/main/kotlin/com/malinskiy/marathon/cli/config/deserialize/AnalyticsConfigurationDeserializer.kt
```diff
@@ -19,6 +19,13 @@ class AnalyticsConfigurationDeserializer : StdDeserializer<AnalyticsConfiguratio
             if (influxDbConfiguration != null) return influxDbConfiguration
         }
 
+        val graphiteNode = node?.get("graphite")
```
Malinskiy: It's possible to specify both configurations at the same time. We should produce an error that it's incorrect to provide both influx and graphite at the same time. Or at the very least, produce a warning that we influx configuration will override the graphite one.

Malinskiy (changes requested):
Good work! I've noticed some things that need some work in the PR and commented on those.
Apart from the requested changes this PR also needs:
- [ ] Integration test with graphite similar to InfluxDbProviderIntegrationTest to verify the initialisation of remote data source
- [ ] Integration test for the data source (similar to InfluxDBDataSourceTest)
- [ ] Add graphite test case in MetricsProviderFactoryTest to test when config is correct and also when it's not correct. Including fallback to NoOp when graphite server is down/unavailable.
- [ ] Add configuration documentation to the `gh-pages` branch in a separate PR so that our users know about this new option.

Also please work on the code coverage until your diff is at least not decreasing the coverage. We can discuss if you'll see some roadblocks on the way, but generally, we don't merge new code that is not well tested.

core/src/main/kotlin/com/malinskiy/marathon/analytics/external/graphite/BasicGraphiteClient.kt
bartek-wesolowski: I copied the code of `BasicGraphiteClient` from [Graphite Kotlin Client](https://github.com/mihkels/graphite-client-kotlin). Unfortunately, the library is not available on Maven Central yet. I'd rather not modify the code, because then it can be easily replaced with the library implementation in the future.

I tried using [graphite-client](https://github.com/savantly-net/graphite-client) before, but for some reason it wasn't sending or fetching any data.

core/src/main/kotlin/com/malinskiy/marathon/analytics/external/MetricsProviderFactory.kt
bartek-wesolowski: `GraphiteDataSource` never throws any exceptions in its constructor. It doesn't create any database connections. Data is fetched on demand via HTTP requests. `createGraphiteMetricsProvider` will always succeed and it doesn't need a fallback.

core/src/main/kotlin/com/malinskiy/marathon/analytics/metrics/remote/graphite/GraphiteDataSource.kt
bartek-wesolowski: Fetching data from Graphite is done via HTTP requests. No database connection is kept open and consequently it doesn't have to be closed. This approach shouldn't slow down tests, because one single query, either for all success rates or for all execution times, is executed before the tests are run and the query takes a fraction of a second.

cli/src/main/kotlin/com/malinskiy/marathon/cli/config/deserialize/GraphiteConfigurationDeserializer.kt
bartek-wesolowski: To be honest, I used `InfluxDbConfigurationDeserializer` as an example and created a similar implementation for Graphite. But yes, it displays detailed error logs when the configuration is incorrect. Do you think this should be handled differently?

cli/src/main/kotlin/com/malinskiy/marathon/cli/config/deserialize/AnalyticsConfigurationDeserializer.kt
bartek-wesolowski: Done. Added throwing a `ConfigurationException` when both InfluxDB and Graphite configurations are specified.

core/src/main/kotlin/com/malinskiy/marathon/analytics/metrics/remote/graphite/GraphiteDataSource.kt
bartek-wesolowski: done

core/src/main/kotlin/com/malinskiy/marathon/analytics/external/graphite/GraphiteTracker.kt
bartek-wesolowski: I tried using tags but they didn't work for me. It turned out that my company disabled them for some technical reason. I'm not able to implement tags right now, because I can't test them. Also, I wouldn't be able to use the Graphite integration if it used tags. I'd rather not change this now and perhaps add tags later as an option.

core/src/main/kotlin/com/malinskiy/marathon/analytics/external/graphite/GraphiteUtils.kt
bartek-wesolowski: done

core/src/main/kotlin/com/malinskiy/marathon/analytics/TrackerFactory.kt
bartek-wesolowski: done

cli/src/main/kotlin/com/malinskiy/marathon/cli/config/deserialize/GraphiteConfigurationDeserializer.kt
bartek-wesolowski: done

bartek-wesolowski:
As to the unit tests:
- IMHO a test similar to `InfluxDbProviderIntegrationTest` is not necessary, because Graphite doesn't have a class like `GraphiteProvider` and `BasicGraphiteClient'`s constructor invocation will never fail.
- I added unit tests for `GraphiteDataSource`
- IMHO additional tests in `MetricsProviderFactoryTest` aren't necessary, because creating `GraphiteDataSource` can never fail

bartek-wesolowski:
Add a PR with documentation:
https://github.com/Malinskiy/marathon/pull/408

bartek-wesolowski:
Could someone please add a "hacktoberfest-accepted" label to this PR so that I can take part in Hacktoberfest 2020 with these changes?

tagantroy:
> * IMHO a test similar to `InfluxDbProviderIntegrationTest` is not necessary, because Graphite doesn't have a class like `GraphiteProvider` and `BasicGraphiteClient'`s constructor invocation will never fail.

I think the point of the integration test - verify the complete integration with a real database. We caught a couple of issues with InfluxDB using this approach.
I think we can merge this diff after the integration test is added.
Thank you for your contribution! :)

bartek-wesolowski:
> > * IMHO a test similar to `InfluxDbProviderIntegrationTest` is not necessary, because Graphite doesn't have a class like `GraphiteProvider` and `BasicGraphiteClient'`s constructor invocation will never fail.
> 
> I think the point of the integration test - verify the complete integration with a real database. We caught a couple of issues with InfluxDB using this approach.
> I think we can merge this diff after the integration test is added.
> Thank you for your contribution! :)

I added another unit test for `BasicGraphiteClient`.

As to the integration test, I don't think I'm able to write it. Opposed to InfluxDB, you can't start a Graphite database from code. The Graphite library that I use just wraps Socket and HTTP connections and it consists of one short file. The unit tests already check whether correct data is sent/fetched.

tagantroy:
May I ask you share some examples of dashboard?(for example https://github.com/Malinskiy/marathon/blob/develop/assets/grafana-dashboard.json) 
I created template for integration test(patch attached)
[patch.txt](https://github.com/Malinskiy/marathon/files/5418714/patch.txt)

bartek-wesolowski:
> May I ask you share some examples of dashboard?(for example https://github.com/Malinskiy/marathon/blob/develop/assets/grafana-dashboard.json)
> I created template for integration test(patch attached)
> [patch.txt](https://github.com/Malinskiy/marathon/files/5418714/patch.txt)

I added a sample dashboard for Graphite. I'll see what I can do about the integration test. The sample dashboard looks like this with some sample data.

Zrzut ekranu 2020-10-23 o 08 18 42

cli/src/main/kotlin/com/malinskiy/marathon/cli/config/deserialize/GraphiteConfigurationDeserializer.kt
Malinskiy: No, that's fine, leave it as is. 
We're thinking of moving to something else in terms of configuration parsing so that we get proper error messages out-of-the-box instead of writing custom deserialisers. The whole point was to define data model and let the parse handle everything, but json/yaml packages lack proper error handling for our particular case.

core/src/main/kotlin/com/malinskiy/marathon/analytics/external/MetricsProviderFactory.kt
Malinskiy: That might be sub-optimal in terms of performance (we might request thousands of metrics throughout the execution of a large test suite). For starters, this should be fine though. No need to change.

core/src/main/kotlin/com/malinskiy/marathon/analytics/external/graphite/BasicGraphiteClient.kt
Malinskiy: gotcha. Can you leave a comment there that this is not original code and leave a link to source? thanks

core/src/main/kotlin/com/malinskiy/marathon/analytics/external/graphite/GraphiteTracker.kt
Malinskiy: They're disabled probably because of high cardinality

core/src/main/kotlin/com/malinskiy/marathon/analytics/external/graphite/BasicGraphiteClient.kt
bartek-wesolowski: done

bartek-wesolowski:
I added an integration test for Graphite and refactored integration tests a little bit so that the test cases are shared between InfluxDb and Graphite tests.

tagantroy (approved):
I'm not a big fan of inheritance in tests, but I think it is ok in this case.
LGTM!

bartek-wesolowski:
@Malinskiy check it out once again please. Your comments have been addressed.

bartek-wesolowski:
Guys, would you mind taking a look at this PR once again? Do you think it can be merged now or does it still need some work?

tagantroy (approved):

PR #415, opened 2020-12-01
Don't try to stop Bugsnag if it's not needed

nimeacuerdo:
Use the same flag already in place for starting (or not) Bugsnag in order to decide if it needs to be stopped too.

### DIFF
--- a/cli/src/main/kotlin/com/malinskiy/marathon/cli/ApplicationView.kt
+++ b/cli/src/main/kotlin/com/malinskiy/marathon/cli/ApplicationView.kt
@@ -59,7 +59,9 @@ fun main(args: Array<String>): Unit = mainBody(
             }
         } finally {
             stopKoin()
-            bugsnagExceptionsReporter.end()
+            if(bugsnagReporting){
+                bugsnagExceptionsReporter.end()
+            }
         }
     }
 }
--- a/cli/src/main/kotlin/com/malinskiy/marathon/cli/ApplicationView.kt
+++ b/cli/src/main/kotlin/com/malinskiy/marathon/cli/ApplicationView.kt
@@ -33,7 +33,7 @@ fun main(args: Array<String>): Unit = mainBody(
         logger.info { "Starting marathon" }
         val bugsnagExceptionsReporter = BugsnagExceptionsReporter()
         try {
-            if(bugsnagReporting){
+            if (bugsnagReporting) {
                 logger.info { "Init BugSnag" }
                 bugsnagExceptionsReporter.start(AppType.CLI)
             }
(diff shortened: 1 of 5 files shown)


cli/src/main/kotlin/com/malinskiy/marathon/cli/ApplicationView.kt
```diff
@@ -59,7 +59,9 @@ fun main(args: Array<String>): Unit = mainBody(
             }
         } finally {
             stopKoin()
-            bugsnagExceptionsReporter.end()
+            if(bugsnagReporting){
```
Malinskiy: We're using the IntelliJ's automatic code style formatting, would you be able to hit the auto-format in the IDE? Alternatively, a space after `if` and after the closing bracket `)`.
Thank you.

Malinskiy (changes requested):

cli/src/main/kotlin/com/malinskiy/marathon/cli/ApplicationView.kt
nimeacuerdo: Sure, I just copied the existing approach from https://github.com/Malinskiy/marathon/blob/063a759ed9d416039936b95a71651fbe9abfa693/cli/src/main/kotlin/com/malinskiy/marathon/cli/ApplicationView.kt#L36, which I guess should be updated too, right?

Malinskiy: Yeah, seems fair. Thanks!

Malinskiy:
The change also needs to be in the gradle plugin: https://github.com/Malinskiy/marathon/blob/develop/marathon-gradle-plugin/src/main/kotlin/com/malinskiy/marathon/MarathonRunTask.kt#L100

I'm leaning towards having this if condition inside the BugsnagExceptionsReporter. Maybe passing the `enabled` configuration option as a constructor parameter and then setting the `private val bugsnag: Bugsnag? by lazy` to be null if it's not enabled. What do you think?

nimeacuerdo:
> The change also needs to be in the gradle plugin: https://github.com/Malinskiy/marathon/blob/develop/marathon-gradle-plugin/src/main/kotlin/com/malinskiy/marathon/MarathonRunTask.kt#L100
> 
> I'm leaning towards having this if condition inside the BugsnagExceptionsReporter. Maybe passing the `enabled` configuration option as a constructor parameter and then setting the `private val bugsnag: Bugsnag? by lazy` to be null if it's not enabled. What do you think?

sure, adding new changes.

Malinskiy (approved):

Malinskiy:
@nimeacuerdo thank you for your contribution 👍

PR #422, opened 2020-12-17
fix(sample): document and add dependency on qameta maven

Malinskiy:


tagantroy (approved):

Malinskiy:
Should not be required once https://github.com/allure-framework/allure-kotlin/issues/49 is resolved

Issue #353, opened 2020-04-23
How to execute tests against only 1 shard of X shards?


ZakTaccardi:
How can I have marathon split up tests in a deterministic way? For example, let’s say a module has 100 tests, I want to split them into 4 different shards, and run tests against one of those shards (so only 25 of the tests)?

previously I did this with Spoon using `--numShards 4 --shard-index 0`

Malinskiy:
Hey @ZakTaccardi

If you want to split the tests into deterministic groups then I suggest marking them with an annotation (something like `@Group1`,`@Group2`) or separating them by package. Then you can use `filteringConfiguration`. Otherwise, you can just specify the FQDN's of the classes.

What's your use-case for this?

ZakTaccardi:
we have the unfortunate limitation of being able to only spin up one emulator per jenkins instance

so to work around this we shard the tests into X groups, and run a separate jenkins job for each group.

Our current workflow with Spoon is:

```
// jenkins job 1
./gradlew lib:spoonDebugAndroidTest -PspoonNumShards=4 -PspoonShardIndex=0
// jenkins job 2
./gradlew lib:spoonDebugAndroidTest -PspoonNumShards=4 -PspoonShardIndex=1
```
And basically spoon just divides the tests automatically for us.

And if we need a 3rd job, we just increase the shard count
```
// jenkins job 1
./gradlew lib:spoonDebugAndroidTest -PspoonNumShards=3 -PspoonShardIndex=0
// jenkins job 2
./gradlew lib:spoonDebugAndroidTest -PspoonNumShards=3 -PspoonShardIndex=1
// jenkins job 3
./gradlew lib:spoonDebugAndroidTest -PspoonNumShards=3 -PspoonShardIndex=2
```

And this guarantees that all tests in `:lib` are executed

Note, we are using https://github.com/jaredsburrows/gradle-spoon-plugin

ZakTaccardi:
> If you want to split the tests into deterministic groups then I suggest marking them with an annotation (something like @Group1,@Group2) or separating them by package. 

This won't work for us because it would require us to modify our code under test

> you can just specify the FQDN's of the classes.

This may work with additional manual effort, but I don't think we would have the ability to get the FQDN's of all the classes at the time we need to inform 

I believe that the `AndroidJUnitRunner` itself supports test sharding via:

https://developer.android.com/training/testing/junit-runner#sharding-tests

[...]

PR #430, opened 2020-12-26
Android platform specific allow and block lists can be specified.

saim80:
At our organization, we use bare metal hosts at cloud as CI test runners. Each machine spawns many emulators. Since single machine can take only so many emulators, we encountered a scalability issue where we need to distribute our tests across multiple CI machines. Thanks to Marathon for being such a stable CI software. But, the default regex based test filter wouldn't really help overcome our scalability issue. I added a filter option only for Android where only designated tests can be run. The input format is a comma-separated list of fully qualified method name which well known on Android.

### DIFF
--- a/cli/src/main/kotlin/com/malinskiy/marathon/cli/args/FileAndroidConfiguration.kt
+++ b/cli/src/main/kotlin/com/malinskiy/marathon/cli/args/FileAndroidConfiguration.kt
@@ -29,7 +29,9 @@ data class FileAndroidConfiguration(
     @JsonProperty("serialStrategy") val serialStrategy: SerialStrategy = SerialStrategy.AUTOMATIC,
     @JsonProperty("screenRecordConfiguration") val screenRecordConfiguration: ScreenRecordConfiguration = ScreenRecordConfiguration(),
     @JsonProperty("waitForDevicesTimeoutMillis") val waitForDevicesTimeoutMillis: Long?,
-    @JsonProperty("allureConfiguration") val allureConfiguration: AllureConfiguration?
+    @JsonProperty("allureConfiguration") val allureConfiguration: AllureConfiguration?,
+    @JsonProperty("allowList") val allowList: String = "",
+    @JsonProperty("blockList") val blockList: String = ""
 ) : FileVendorConfiguration {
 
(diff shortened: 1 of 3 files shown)


tagantroy:
Hi, what is the difference with existing allow/block lists?
https://github.com/Malinskiy/marathon/blob/develop/cli/src/test/resources/fixture/config/sample_1.yaml#L33

saim80:
Our system has test distributor to support multiple macOS machine which launches about 9 emulators. And, we have a list of test in the format of "package.classA#methodA,package.classA#methodB' from the distributor. 

For example,
package.classA.methodA method can be given to the mac machine 'A'
packcge.classA.methodB method can be given to the mac machine 'B'

For *exact* matches, the current system offers "composition" filter type. where I can specify everything with "INTERSECTION" op. And, I had two problems.

1. Too verbose. The Marathonfile gets way too big and unreadable as I update it through a script. Imagine the repetition for this exact match expression.
2. Slow. please correct me if I'm wrong, but regex filter is being applied to each test cases. This means, for the exact match filter, if I have 'A' filters in Marathonfile and 'B' test cases. It will take 'AB' times to finish all the filterings.  Again, the number 'A' is the number of test cases given by the external test distributor.

This was just a quick mitigation for myself. But, if there can be a better option for exact match filters, I can migrate.

Malinskiy:
This sounds to me like a manual version of #353 where the shards are prefilled by the user.

From the user perspective I think regex vs list use-case makes sense. We should start supporting a list of tests either by an actual list in the yaml or a completely separate file, e.g.:
Marathonfile:
```yaml
filteringConfiguration:
  file: external.yaml
```

and the external.yaml will look something like:
```
allow:
- com.example.MyTest
block:
- com.example.MyTest2
```

This option could also be overridden from CLI since the use-case resuses the Marathonfile and it's essentially the same run.

What I don't like is that we support the distributed hardware in marathon but we don't use that and fall back to stone-age sharding. I'd like to ask @saim80 if you have network connectivity between your CI agents? Would you be able to connect via TCP/IP from one agent to another one?

Malinskiy:
@saim80 bump

Malinskiy:
Closing due to inactivity. Feel free to reopen with the answers to our questions

PR #426, opened 2020-12-23
Stabilise 0.6.0 adam

Malinskiy:
This PR stabilizes the adam Android vendor. List of issues:
* Adam 0.2.3 (Fix socket timeout for long running requests + USB transport)
* Force single kotlin version everywhere
* Update gradle to 6.8 and kotlin to 1.4.20
* Fix for suspend race condition in QueueActor
* Use new image capture adapters from adam
* New timeout configuration (adam specific)
* Measure file transfer speeds during pull/push
* Survive adb kill-server
* Fix allure legacy storage on API 29
* Fix AndroidAppInstaller for some devices
* Rotate screenshots based on initial device rotation

PR #439, opened 2021-02-01
Fix/uncompleted tests

### LINKED ISSUE #428: Uncompleted tests are not reported as failed after max retry count
**Describe the bug**
If some test retries the maximum amount of times specified by the retry strategy, the test is completely missing a failed status

**To Reproduce**
Steps to reproduce the behaviour:
1. Use the sample android-app
2. Increase one test delay over the batch limit

**Expected behavior**
Test fails after the max retry count even if it was never reported

**Logs and reports**

**Devices (please complete the following information):**

**Additional context**

Malinskiy:
Uncompleted tests should not be counted with retryStrategy. The only parameter that limits the amount should be `uncompletedTestRetryQuota`. After all the attempts to retry the INCOMPLETE test failed, the marathon will report the test as FAILED.

Previously the test would not end up in the reports as failed, this behaviour will likely surface tests that were never able to finish before.

Fixes #428 

Depends on #426


vendor/vendor-android/base/src/main/kotlin/com/malinskiy/marathon/android/executor/listeners/TestRunResultsListener.kt
```diff
@@ -63,11 +63,17 @@ class TestRunResultsListener(
             results[it.test.identifier()]?.isSuccessful() ?: false
         }
 
-        val failed = nonNullTestResults.filterNot {
-            results[it.test.identifier()]?.isSuccessful() ?: false
+        val (reportedIncompleteTests, reportedNonNullTests) = nonNullTestResults.partition { it.status == TestStatus.INCOMPLETE }
+
+        val failed = reportedNonNullTests.filterNot {
+            val status = results[it.test.identifier()]
+            when {
```
tagantroy, line 70: same as `status?.isSuccessful() ?: false` ?

tagantroy (approved):

PR #444, opened 2021-02-03
feat(android): support pulling folders after test run

### LINKED ISSUE #392: Add support for taking screenshots (Android)
I'm not seeing any documentation related to taking screenshots on Android with Marathon. What is the right way to do this?

Malinskiy:
This breaks the location of the allure device results, but it's better we do it now rather than later. The current approach will allow
us to pull coverage, arbitrary screenshot frameworks and more


Malinskiy:
Resolves #392 for pulling random screenshot folders from the device

Vacxe (approved):
Tested with #466

Issue #442, opened 2021-02-02
Migrate gradle plugin to a simple wrapper around marathon CLI


Malinskiy:
Due to the increasing amount of issues with classpath in the gradle plugin it's reasonable to rewrite it to:
1. Generate Marathonfile
2. Call marathon cli in a separate JVM

Malinskiy:
Closed with #573

Issue #413, opened 2020-10-31
Migrate CLI configuration to hoplite


tagantroy:
Migrate CLI configuration to [Hoplite](https://github.com/sksamuel/hoplite)
It supports data/sealed classes, optional values, tuples and etc. [link](https://github.com/sksamuel/hoplite#decoders)
[Supported formats](https://github.com/sksamuel/hoplite#supported-formats): yaml, toml, json and etc. It also support env variables in yaml.

sphanley:
Asking this with nothing but appreciation for the hard work of the team here: anything you're able to share in terms of an ETA for this feature? My team is working on adding capabilities to our testing design where this would be hugely beneficial (around the ability to dynamically specify subsets of tests to execute), and knowing roughly how soon this is expected would make a big difference in terms of the work we decide to do in that area in the meantime.

PR #448, opened 2021-02-07
Create cli configuration module

tagantroy:
Add abstraction layer for everything related to Marathonfile configuration. 
This module will be used for #442 and #413


tagantroy:
Will be merged as part of new plugin and cli

Issue #440, opened 2021-02-01
Change Android Vendor to ADAM in Gradle configuration


asos-jorgegonzalez:
I want to switch to ADAM Vendor using the Gradle configurationo but I can't achieve it due to this error: 
```
Caused by: groovy.lang.MissingPropertyException: Could not get unknown property 'com' for extension 'marathon' of type com.malinskiy.marathon.MarathonExtension.
```
How can I import the vendor into the Marathon config? What am I missing?

My files:

**root/settings.gradle**
```
pluginManagement {
    repositories {
        google()
        mavenCentral()
        maven { url 'https://dl.bintray.com/qameta/maven' }
        gradlePluginPortal()
    }
    resolutionStrategy {
        eachPlugin {
            if (requested.id.id == "marathon") {
                useModule("com.malinskiy.marathon:marathon-gradle-plugin:0.6.0")
            }
        }
    }
}

[...]
```

**root/build.gradle**
```
buildscript {
    repositories {
        [...]
    }
    dependencies {
        classpath "com.malinskiy.marathon:marathon-gradle-plugin:0.6.0"

        [...]        
    }
}

[...]
```

**root/app/build.gradle**
```
apply plugin: 'marathon'

marathon {

    vendor = com.malinskiy.marathon.android.VendorType.ADAM // <----

    [...]
}
```

Malinskiy:
Hey @asos-jorgegonzalez 
I'll check this today or tomorrow and will get back to you

My only suspicion is that you might need an import statement since gradle interprets this as a property rather than FQDN

asos-jorgegonzalez:
Thanks! Is it possible that the cause is that Vendor type is declared in a different module than `marathon-gradle-plugin`? Then we don't have access to it. The same issue happens if you want to use ScreenRecordConfigurations. I tried to add those maven modules into **root/build.gradle**, but it doesn't work.
```
buildscript {
    repositories {
        [...]
    }
    dependencies {
        classpath "com.malinskiy.marathon:marathon-gradle-plugin:0.6.0"
        classpath "com.malinskiy.marathon:base:0.6.0"
        classpath "com.malinskiy.marathon:core:0.6.0"

        [...]        
    }
}
```

Could a solution be to replace `implementation` with `api` for those modules inside **marathon-gradle-plugin/build.gradle.kts**? I mean:

**marathon-gradle-plugin/build.gradle.kts**
```
[...]

PR #457, opened 2021-02-17
Feature (Gradle): add serialStrategyName property

snpefk:
At my organization we have a project with a structure something like this: 
```
📁 .
├── 📁 androidModuleA
│  └── 💾 build.gradle
├── 📁 androidModuleB
│  └── 💾 build.gradle
├── 📁 androidModuleC
│  └── 💾 build.gradle
├── 💾 build.gradle
└── 💾 marathon.gradle
```

Each submodule applies `marathon.gradle` in its own `build.gradle` file so solution from #440 does not suit for us. It's not very convenient for us to copy-paste same thing. 

So, I suggest adding `serialStrategyName` property (working title) that accept string to eliminate the need to import `SerialStrategy` enum and avoid breaking change by changing `serialStrategy` type.


Malinskiy:
cc @tagantroy

Malinskiy:
@snpefk can you be a bit more specific in terms of your solution? Please paste the code relevant to the question.

clarification: the solution for using the imported ADAM enum value, not the PR that you have here

snpefk:
@Malinskiy Hi, sorry for late reply. 

Solution we tried is to copy `marathon` block from `marathon.gradle` file to several `build.gradle` files in submodules: 
```gradle
// marathon.gradle
marathon {
    applicationPmClear = true
    testApplicationPmClear = true
    …
    // this is where problem arose  
    // Caused by: groovy.lang.MissingPropertyException: Could not get unknown property 'com' for extension 'marathon'
    serialStrategy = com.malinskiy.marathon.android.serial.SerialStrategy.BOOT_PROPERTY
}
```
I cut this block and paste to `build.gradle` submodules (this is roughly how all `build.gradle` files looks now): 
```gradle
// androidModuleX/build.gralde
android { 
    buildTypes {
        …
    }
    productFlavors {
        …
    }
}
 // and now there marathon block 
marathon {
     // no more error    
    serialStrategy = com.malinskiy.marathon.android.serial.SerialStrategy.BOOT_PROPERTY
}
```

In #440 you suggest to copy `marathon` block to
> Wherever you apply the application/library android plugin - the marathon apply block should be also there

Although this was a different case solution is still relevant (and nothing else I tried work). 

(I hope I understand your question correctly)

PR #459, opened 2021-02-19
Created annotation data filter for filtering by value in annotation

SergKhram:


core/src/main/kotlin/com/malinskiy/marathon/execution/Filter.kt
```diff
@@ -87,3 +87,25 @@ data class TestMethodFilter(@JsonProperty("regex") val regex: Regex) : TestFilte
         return "TestMethodFilter(regex=$regex)"
     }
 }
+
+data class AnnotationDataFilter(@JsonProperty("nameRegex") val nameRegex: Regex, @JsonProperty("valueRegex") val valueRegex: Regex ) : TestFilter {
+    override fun filter(tests: List<Test>): List<Test> = tests.filter {
+        it.metaProperties.map { it.name }.any(nameRegex::matches) &&
```
tagantroy: May I ask you add unit test to verify use case with two annotations? 
For example:
We have a test with two annotations.
```
val test = stubTest(AnnotationData("com.example.CorrectAnnotation", "INVALID_VALUE"),
                              AnnotationData("com.example.IncorrectAnnotation", "CORRECT_VALUE")
)
```
And we want to find tests with com.example.CorrectAnnotation and CORRECT_VALUE.
```
val filter = AnnotationDataFilter("""com\.example\.CorrectAnnotation""".toRegex(), "CORRECT_VALUE".toRegex())
```
Result in this case should be emptyList(), because we cannot find pair of com.example.CorrectAnnotation + CORRECT_VALUE

tagantroy (commented):
Thanks for contribution, 
Please take a look on my comment about filter/filterNot implementation. 
It would be nice to see this test case.

core/src/main/kotlin/com/malinskiy/marathon/execution/Filter.kt
SergKhram: Done

tagantroy (approved):

PR #463, opened 2021-02-26
Ktor socketTimeout config option

trevjonez:
What I did to help diagnose https://github.com/Malinskiy/marathon/issues/462

### DIFF
--- a/vendor/vendor-android/base/src/main/kotlin/com/malinskiy/marathon/android/configuration/TimeoutConfiguration.kt
+++ b/vendor/vendor-android/base/src/main/kotlin/com/malinskiy/marathon/android/configuration/TimeoutConfiguration.kt
@@ -12,4 +12,5 @@ data class TimeoutConfiguration(
     var install: Duration = shell,
     var screenrecorder: Duration = Duration.ofMinutes(10),
     var screencapturer: Duration = Duration.ofMillis(300),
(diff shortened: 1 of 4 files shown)


vendor/vendor-android/base/src/main/kotlin/com/malinskiy/marathon/android/configuration/TimeoutConfiguration.kt
```diff
@@ -12,4 +12,5 @@ data class TimeoutConfiguration(
     var install: Duration = shell,
     var screenrecorder: Duration = Duration.ofMinutes(10),
     var screencapturer: Duration = Duration.ofMillis(300),
+    var socketTimeout: Duration = Duration.ofSeconds(30)
```
Malinskiy: @trevjonez Can we rename this to `socketIdleTimeout`? Sockets have more timeout parameters, ktor just doesn't expose it currently

Malinskiy (changes requested):

vendor/vendor-android/base/src/main/kotlin/com/malinskiy/marathon/android/configuration/TimeoutConfiguration.kt
trevjonez: very valid point, done.

PR #475, opened 2021-04-01
Caught IllegalStateException on ADB init in case of terminated previous run

SergKhram:
If in the previous run we have some troubles and as the result we terminated test run -> in the current test run we have this error:
`> Task :app:marathonDebugAndroidTest FAILED
ddmlib Android vendor will be deprecated in 0.7.0 and is scheduled to be removed in 0.8.0.
        More info: https://malinskiy.github.io/marathon/ven/android.html#vendor-module-selection
java.lang.IllegalStateException: AndroidDebugBridge.init() has already been called.
Finish BugSnag

FAILURE: Build failed with an exception.`


SergKhram:
@tagantroy check it pls)

tagantroy:
I think better solution here is handling close correctly for resources. We can land this change as a workaround. @Malinskiy what do you think?

Malinskiy:
I agree with @tagantroy here, we should handle the termination properly.
The underlying problem here is that terminating Gradle worker task doesn't close the ddmlib bridge, so we should try to handle this appropriately there instead of catching the exception like this.

SergKhram:
@Malinskiy @tagantroy I tried to use ShutdownHooks.addShutdownHook(PluginShutdownHook) but it isn't work
`class PluginShutdownHook : Thread() {
    override fun run() {
        AndroidDebugBridge.terminate()
    }
}`

I don't understand where I need to paste it

i tried two places in MarathonPlugin.kt:
inside marathonTask = project.task (45 line)
inside marathonTask.configure(93 line) 


After that i rewrite logic in current commit, check it pls

tagantroy:
It will be fixed in #573

Malinskiy:
Closing due to #573

PR #479, opened 2021-04-23
add property annotations to MarathonRunTask

RBusarow:
fixes #477

Issue #477, opened 2021-04-14
Type 'MarathonRunTask' property 'applicationVariant' is missing an input or output annotation


dcampogiani:
**Describe the bug**
I'm getting 
` Type 'MarathonRunTask' property 'applicationVariant' is missing an input or output annotation`
when running `marathonDebugAndroidTest` with:
 - Marathon Gradle Plugin 0.6.1
 - Gradle 7.0

**To Reproduce**
1. Use Gradle 7.0
2. Apply marathon Gradle plugin 0.6.1
3. Run marathonDebugAndroidTest task

**Expected behavior**
marathonDebugAndroidTest should be executed

**Logs and reports**
```
Some problems were found with the configuration of task ':app:marathonDebugAndroidTest' (type 'MarathonRunTask').
  - Type 'MarathonRunTask' property 'applicationVariant' is missing an input or output annotation.
```

**Additional context**
Everything is fine with the Gradle 6.x release

PR #481, opened 2021-04-25
Fix MarathonRunTask inputs #477

tagantroy:
Replace @Input to @Internal for applicationVariant, testVariant, extensionConfig and exceptionsTracker.
Fix problem introduced in  #479
```
lateinit property cnf has not been initialized
```
```
Unable to store input properties for task ':app:marathonDebugAndroidTest'. Property 'extensionConfig' with value 'extension 'marathon'' cannot be serialized.
```
```
Unable to store input properties for task ':app:marathonDebugAndroidTest'. Property 'applicationVariant' with value 'com.android.build.gradle.internal.api.ApplicationVariantImpl_Decorated@5b9982d7' cannot be serialized.
```

```
Unable to store input properties for task ':app:marathonDebugAndroidTest'. Property 'testVariant' with value 'com.android.build.gradle.internal.api.TestVariantImpl_Decorated@3aefcb61' cannot be serialized.
```

```
Unable to store input properties for task ':app:marathonDebugAndroidTest'. Property 'exceptionsTracker' with value 'com.malinskiy.marathon.exceptions.BugsnagExceptionsReporter@1e79d8d9' cannot be serialized.
```

### DIFF
--- a/marathon-gradle-plugin/src/main/kotlin/com/malinskiy/marathon/MarathonRunTask.kt
+++ b/marathon-gradle-plugin/src/main/kotlin/com/malinskiy/marathon/MarathonRunTask.kt
@@ -25,49 +25,61 @@ import com.malinskiy.marathon.usageanalytics.tracker.Event
 import ddmlibModule
 import org.gradle.api.DefaultTask
 import org.gradle.api.GradleException
+import org.gradle.api.file.DirectoryProperty
+import org.gradle.api.model.ObjectFactory
+import org.gradle.api.provider.Property
 import org.gradle.api.tasks.Input
 import org.gradle.api.tasks.InputDirectory
+import org.gradle.api.tasks.Internal
 import org.gradle.api.tasks.OutputDirectory
+import org.gradle.api.tasks.PathSensitive
+import org.gradle.api.tasks.PathSensitivity
 import org.gradle.api.tasks.TaskAction
 import org.gradle.api.tasks.VerificationTask
+import org.gradle.kotlin.dsl.property
 import org.koin.core.context.stopKoin
 import java.io.File
 
 private val log = MarathonLogging.logger {}
 
-open class MarathonRunTask : DefaultTask(), VerificationTask {
+open class MarathonRunTask(objects: ObjectFactory) : DefaultTask(), VerificationTask {
     @Input
-    lateinit var flavorName: String
+    val flavorName: Property<String> = objects.property()
+
     @Input
-    lateinit var applicationVariant: BaseVariant
+    val applicationVariant: Property<BaseVariant> = objects.property()
+
     @Input
-    lateinit var testVariant: TestVariant
-    @Input
-    lateinit var extensionConfig: MarathonExtension
+    val testVariant: Property<TestVariant> = objects.property()
+
     @InputDirectory
-    lateinit var sdk: File
+    @PathSensitive(PathSensitivity.NAME_ONLY)
+    val sdk: DirectoryProperty = objects.directoryProperty()
+
     @Input
-    lateinit var cnf: Configuration
-    @Input
-    lateinit var exceptionsTracker: ExceptionsReporter
-    @Input
-    var ignoreFailure: Boolean = false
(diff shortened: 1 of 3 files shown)


marathon-gradle-plugin/src/main/kotlin/com/malinskiy/marathon/MarathonRunTask.kt
```diff
@@ -38,17 +39,15 @@ private val log = MarathonLogging.logger {}
 open class MarathonRunTask : DefaultTask(), VerificationTask {
     @Input
     lateinit var flavorName: String
```
sergioboi: Does the `flavorName` make sense as input when everything else is internal?

marathon-gradle-plugin/src/main/kotlin/com/malinskiy/marathon/MarathonRunTask.kt
```diff
@@ -38,17 +39,15 @@ private val log = MarathonLogging.logger {}
...
-    @Input
+    @Internal
     lateinit var applicationVariant: BaseVariant
```
sergioboi: You should prefer `Property` to `lateinit`
Property connects with Gradle lifecycle and has greater API

marathon-gradle-plugin/src/main/kotlin/com/malinskiy/marathon/MarathonRunTask.kt
```diff
@@ -38,17 +39,15 @@ private val log = MarathonLogging.logger {}
...
     @InputDirectory
     lateinit var sdk: File
-    @Input
-    lateinit var cnf: Configuration
-    @Input
+    @Internal
+    lateinit var extensionConfig: MarathonExtension
+    @Internal
     lateinit var exceptionsTracker: ExceptionsReporter
     @Input
```
sergioboi, line 59: The same about that `Input`
You should rethink what fields make sense to be input because now it could be hard for clients to realize why their tasks are UP-TO-DATE

Malinskiy:
`@Internal` doesn't make sense for us here. Maybe for the exceptions tracker. 

In general:

1. flavorName should be considered for up-to-date checking, otherwise different flavors won't be tested separately
2. applicationVariant - same as 1
3. testVariant: TestVariant
4. sdk: if it changes we should be okay-ish? I'd still consider adding it to Input
5. extensionConfig: MarathonExtension - this is the main input
6. exceptionsTracker - we should be fine if it changes, no need to invalidate anything
7. ignoreFailure - part of configuration

tagantroy:
> `@Internal` doesn't make sense for us here. Maybe for the exceptions tracker.
> 
> In general:
> 
>     1. flavorName should be considered for up-to-date checking, otherwise different flavors won't be tested separately
> 
>     2. applicationVariant - same as 1
> 
>     3. testVariant: TestVariant
> 
>     4. sdk: if it changes we should be okay-ish? I'd still consider adding it to Input
> 
>     5. extensionConfig: MarathonExtension - this is the main input
> 
>     6. exceptionsTracker - we should be fine if it changes, no need to invalidate anything
> 
>     7. ignoreFailure - part of configuration

I don't see any difference in our case to be honest. We always set `outputs.upToDateWhen { false }` and it prevents task from caching. Changed all fields(except exceptionsTracker) to Property<T> with @Input.

marathon-gradle-plugin/src/main/kotlin/com/malinskiy/marathon/MarathonRunTask.kt
tagantroy: All fields are migrated.

marathon-gradle-plugin/src/main/kotlin/com/malinskiy/marathon/MarathonRunTask.kt
tagantroy: Caching is actually disabled with outputs.upToDateWhen { false }

Malinskiy (approved):

PR #484, opened 2021-04-25
[Gradle 7.0 migration] Upgrade gradle wrapper

tagantroy:
Use latest Gradle version.
This PR depends on #483


Malinskiy:
Should probably contain updates to samples as well?

tagantroy:
> Should probably contain updates to samples as well?

I'll update  sample apps in separate diff. It will require some extra work on gradle plugin.

buildSrc/src/main/kotlin/Versions.kt
```diff
@@ -1,7 +1,7 @@
 object Versions {
     val marathon = System.getenv("DEPLOY_VERSION_OVERRIDE") ?: "0.7.0"
 
-    val kotlin = "1.4.20"
+    val kotlin = "1.4.31"
```
Vacxe, line 4: 1.4.32 - latest stable

Vacxe (approved):

buildSrc/src/main/kotlin/Versions.kt
tagantroy: I checked release notes of Gradle 7.0, they upgraded Kotlin to 1.4.31(
[#16330] - Upgrade embedded Kotlin to 1.4.31). I hope upgrade to 1.4.32 will not cause any issue, but I would prefer to keep version same. Let's check it later

PR #488, opened 2021-04-28
[WIP][Gradle 7.0][AGP 4.1] Migrate gradle plugin to AGP 4.1 API, upgrade gradle version for sample apps

tagantroy:
AGP 4.1 released on October 12, 2020. https://androidstudio.googleblog.com/2020/10/android-studio-41-available-in-stable.html
I think it is time for migration to the new API. 
Use Property<T> for extension fields.
Use Action<T> instead of Closure for DSL configuration.


marathon-gradle-plugin/src/main/kotlin/com/malinskiy/marathon/MarathonExtension.kt
```diff
@@ -7,183 +7,120 @@ import com.malinskiy.marathon.android.configuration.FileSyncConfiguration
...
+import org.gradle.api.provider.MapProperty
+import org.gradle.api.provider.Property
+import org.gradle.kotlin.dsl.mapProperty
+import org.gradle.kotlin.dsl.property
+import java.io.Serializable
+import javax.inject.Inject
 
-open class MarathonExtension(project: Project) {
-    var name: String = "Marathon"
+open class MarathonExtension @Inject constructor(objects: ObjectFactory) : Serializable {
```
sergioboi, line 19: There is one more trick
If you make a class `abstract` and mark property `abstract` Gradle will initialize them for you
It helps to reduce the amount of code

marathon-gradle-plugin/src/main/kotlin/com/malinskiy/marathon/MarathonPlugin.kt
```diff
@@ -1,136 +1,112 @@
...
+                        timeoutConfiguration.set(extension.timeoutConfiguration)
+                        fileSyncConfiguration.set(extension.fileSyncConfiguration)
+
+                        //Android specific for now
+                        autoGrantPermission.set(extension.autoGrantPermission)
+                        instrumentationArgs.set(extension.instrumentationArgs)
+
+
+                        sdk.set(androidExtension.sdkComponents.sdkDirectory)
+                        outputs.upToDateWhen { false }
```
plastiv, line 105: nit: I think `outputs.cacheIf { false }` should go in pair with `outputs.upToDateWhen { false }` to prevent upload cached artifacts which would never be reused.

PR #499, opened 2021-05-13
feat(core): test retry artifacts

Malinskiy:
This effectively saves all the failed test artifacts for default mode of execution: even if a test succeeded, you will get failed retries to investigate flaky tests.

Retry screen recordings + logs will be visible via allure report as well as available via raw file access

### DIFF
--- a/core/src/integrationTest/kotlin/com/malinskiy/marathon/analytics/metrics/remote/TestEvents.kt
+++ b/core/src/integrationTest/kotlin/com/malinskiy/marathon/analytics/metrics/remote/TestEvents.kt
@@ -10,6 +10,7 @@ import com.malinskiy.marathon.execution.TestStatus
 import com.malinskiy.marathon.generateTest
 import java.time.Instant
 import java.time.temporal.ChronoUnit
+import java.util.*
 
 val test = generateTest()
 
@@ -89,7 +90,8 @@ private fun creteTestEvent(device: DeviceInfo, duration: Long, status: TestStatu
         device = device,
         status = status,
         startTime = whenWasSent.minusMillis(duration).toEpochMilli(),
-        endTime = whenWasSent.toEpochMilli()
+        endTime = whenWasSent.toEpochMilli(),
+        testBatchId = UUID.randomUUID().toString()
     ),
     true
 )
--- a/core/src/integrationTest/kotlin/com/malinskiy/marathon/analytics/metrics/remote/TestEvents.kt
+++ b/core/src/integrationTest/kotlin/com/malinskiy/marathon/analytics/metrics/remote/TestEvents.kt
@@ -10,7 +10,7 @@ import com.malinskiy.marathon.execution.TestStatus
 import com.malinskiy.marathon.generateTest
 import java.time.Instant
 import java.time.temporal.ChronoUnit
-import java.util.*
+import java.util.UUID
 
 val test = generateTest()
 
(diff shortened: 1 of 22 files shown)


core/src/integrationTest/kotlin/com/malinskiy/marathon/analytics/metrics/remote/TestEvents.kt
```diff
@@ -10,6 +10,7 @@ import com.malinskiy.marathon.execution.TestStatus
 import com.malinskiy.marathon.generateTest
 import java.time.Instant
 import java.time.temporal.ChronoUnit
+import java.util.*
```
sergioboi: Maybe add detekt and suppress `star imports`

core/src/main/kotlin/com/malinskiy/marathon/io/FileManager.kt
```diff
@@ -11,9 +11,9 @@ import java.nio.file.Paths.get
 
 @Suppress("TooManyFunctions")
 class FileManager(private val output: File) {
-    fun createFile(fileType: FileType, pool: DevicePoolId, device: DeviceInfo, test: Test): File {
+    fun createFile(fileType: FileType, pool: DevicePoolId, device: DeviceInfo, test: Test, testBatchId: String? = null): File {
```
sergioboi, line 14: Why do we need a default value for `testBatchId` param?

core/src/main/kotlin/com/malinskiy/marathon/report/html/HtmlSummaryReporter.kt
```diff
@@ -154,34 +160,34 @@ class HtmlSummaryReporter(
         modelName = model
     )
 
-    private fun TestResult.receiveScreenshotPath(poolId: String): String {
-        val screenshotRelativePath = "/screenshot/$poolId/${device.serialNumber}/${test.pkg}.${test.clazz}#${test.method}.gif"
+    private fun TestResult.receiveScreenshotPath(poolId: String, batchId: String): String {
```
Vacxe, line 163: Can be `batchId` is empty in this case and similar below?

core/src/integrationTest/kotlin/com/malinskiy/marathon/analytics/metrics/remote/TestEvents.kt
Malinskiy: detekt is enabled, we ignore it in the CI
IDEA codestyle is better, will check why it hasn't taken effect here

core/src/main/kotlin/com/malinskiy/marathon/io/FileManager.kt
Malinskiy: Some output files are not associated with a batch, e.g. test `json`

core/src/main/kotlin/com/malinskiy/marathon/report/html/HtmlSummaryReporter.kt
Malinskiy: No, it can't be empty - you have to provide it everywhere for screencapture + log files, hence no default value.
Or do you refer to the validation logic?

core/src/integrationTest/kotlin/com/malinskiy/marathon/analytics/metrics/remote/TestEvents.kt
Malinskiy: fixed the codestyle autoformatting and the import

tagantroy (approved):

PR #510, opened 2021-05-30
Add strict mode fail fast

sergioboi:
This is a first part where we have to realize at the Device level that we have a failure


cli/src/main/kotlin/com/malinskiy/marathon/cli/config/deserialize/StrictModeDeserializer.kt
```diff
@@ -0,0 +1,16 @@
...
+import com.fasterxml.jackson.databind.DeserializationContext
+import com.fasterxml.jackson.databind.deser.std.StdDeserializer
+import com.malinskiy.marathon.execution.strategy.StrictMode
+
+class StrictModeDeserializer : StdDeserializer<StrictMode>(StrictMode::class.java) {
+    override fun deserialize(
+        p: JsonParser,
+        ctxt: DeserializationContext
+    ): StrictMode {
+        val value = p.valueAsString.toUpperCase()
```
sergioboi: Need to check what user will get in console

core/src/main/kotlin/com/malinskiy/marathon/execution/TestBatchResults.kt
```diff
@@ -6,5 +6,14 @@ data class TestBatchResults(
     val device: Device,
     val finished: Collection<TestResult>,
     val failed: Collection<TestResult>,
-    val uncompleted: Collection<TestResult>
-)
+    val uncompleted: Collection<TestResult>,
+    val runCompletionReason: RunCompletionReason
```
sergioboi, line 10: We discussed that TestBatchResult should have info about the reason why it completes.
Maybe you have better naming purposes

core/src/main/kotlin/com/malinskiy/marathon/execution/queue/TestResultReporter.kt
```diff
@@ -57,14 +68,21 @@ class TestResultReporter(
...
+                            transitionTo(TestState.Executed(it.device, it.testResult, count - 1))
+                        StrictMode.ALL_SUCCESS,
+                        StrictMode.ANY_FAIL ->
+                            transitionTo(TestState.Failed(it.device, it.testResult), TestAction.SaveReport(it.device, it.testResult))
+                    }
                 }
             }
             on<TestEvent.Incomplete> {
                 if (it.final) {
+                    // TODO
```
sergioboi: How should we react to Incomplete test when ANY_FAIL was set?

vendor/vendor-android/adam/src/main/kotlin/com/malinskiy/marathon/android/adam/AndroidDeviceTestRunner.kt
```diff
@@ -68,14 +70,25 @@ class AndroidDeviceTestRunner(private val device: AdamAndroidDevice) {
                             return@withTimeoutOrNull
                         } else {
                             processEvents(update, listener)
+                            when (configuration.strictMode) {
```
sergioboi: @Malinskiy We've discussed that the better way is having this logic in a listener but I can't `return` from listener

vendor/vendor-ios/src/main/kotlin/com/malinskiy/marathon/ios/logparser/listener/ProgressReportingListener.kt
```diff
@@ -34,7 +35,7 @@ class ProgressReportingListener(
             !receivedTestNames.contains(it.toSafeTestName())
         }.createUncompletedTestResults(received)
 
-        deferredResults.complete(TestBatchResults(device, success, failure, uncompleted))
+        deferredResults.complete(TestBatchResults(device, success, failure, uncompleted, RUN_END))
```
sergioboi, line 38: I have to touch ios part because we are reusing `TestBatchResults`

.gitignore
```diff
@@ -275,4 +275,19 @@ DerivedData
...
+# Android Sdk
+
+local.properties
+
+## IDEA specific
+.idea/markdown-navigator.xml
+.idea/markdown-navigator-enh.xml
+.idea/misc.xml
+.idea/modules.xml
+.idea/runConfigurations.xml
```
Malinskiy: We do share runConfigurations so that you can import the project and can run a debug session for sample-app. We shouldn't ignore it

Not sure about the other ones

cli/src/main/kotlin/com/malinskiy/marathon/cli/config/deserialize/StrictModeDeserializer.kt
Malinskiy: This deserializer has to be compatible with the old version: false + true and print a deprecation message in such case

cli/src/test/resources/fixture/config/sample_1_rp.yaml
```diff
@@ -65,7 +65,7 @@ excludeSerialRegexes:
 ignoreFailures: false
 isCodeCoverageEnabled: false
 fallbackToScreenshots: false
-strictMode: true
+strictMode: all_success
```
Malinskiy: this test still has to be the same and use the deprecated mode for deserializer

core/src/main/kotlin/com/malinskiy/marathon/execution/TestBatchResults.kt
```diff
@@ -6,5 +6,14 @@ data class TestBatchResults(
...
+) {
+
+    enum class RunCompletionReason {
+        RUN_FAILED,
+        RUN_STOPPED,
```
Malinskiy, line 15: Stopped in our case also includes timeout

core/src/main/kotlin/com/malinskiy/marathon/execution/queue/TestResultReporter.kt
Malinskiy: We should restart it since it might be a device problem. The final flag is indicating if we reached uncompleted retry quota for example

core/src/main/kotlin/com/malinskiy/marathon/execution/strategy/StrictMode.kt
```diff
@@ -0,0 +1,7 @@
+package com.malinskiy.marathon.execution.strategy
+
+enum class StrictMode {
```
Malinskiy: In a sense this is not a strict mode anymore, it's a way to treat failures in the build. Not sure if there is a better naming

vendor/vendor-android/adam/src/main/kotlin/com/malinskiy/marathon/android/adam/AndroidDeviceTestRunner.kt
Malinskiy: You should also close the request here

vendor/vendor-android/base/src/main/kotlin/com/malinskiy/marathon/android/executor/listeners/AndroidTestRunListener.kt
```diff
@@ -3,23 +3,25 @@ package com.malinskiy.marathon.android.executor.listeners
...
+    suspend fun testEnded(test: TestIdentifier, testMetrics: Map<String, String>)
 
-    suspend fun testRunFailed(errorMessage: String) {}
+    suspend fun testRunFailed(errorMessage: String)
 
-    suspend fun testRunStopped(elapsedTime: Long) {}
+    suspend fun testRunStopped(elapsedTime: Long)
 
-    suspend fun testRunEnded(elapsedTime: Long, runMetrics: Map<String, String>) {}
+    suspend fun testRunStoppedFailFast(elapsedTime: Long)
```
Malinskiy: Why do we need an additional method for this particular case?

Malinskiy (changes requested):

core/src/main/kotlin/com/malinskiy/marathon/execution/queue/TestResultReporter.kt
sergioboi: So, I could change nothing in this state

vendor/vendor-android/adam/src/main/kotlin/com/malinskiy/marathon/android/adam/AndroidDeviceTestRunner.kt
sergioboi: As I understand we have `finally` block in that `channel` will be closed when we `return`

core/src/main/kotlin/com/malinskiy/marathon/execution/strategy/StrictMode.kt
sergioboi: I think the better could be `execution_strategy`

vendor/vendor-android/base/src/main/kotlin/com/malinskiy/marathon/android/executor/listeners/AndroidTestRunListener.kt
sergioboi: I make this for the cleanliness of my changes. So, should I use `testRunStopped`?

core/src/main/kotlin/com/malinskiy/marathon/execution/TestBatchResults.kt
sergioboi: As I understand I could delete `RUN_STOPPED_FAIL_FAST` and should use `RUN_STOPPED`

cli/src/main/kotlin/com/malinskiy/marathon/cli/config/deserialize/StrictModeDeserializer.kt
```diff
@@ -0,0 +1,46 @@
...
+        }
+    }
+
+    private fun parseDeprecatedStrictMode(valueAsString: String): StrictMode {
+        logger.warn(
+            """
+            |You use deprecated strict_mode configuration: [true, false].
+            |This API will be deleted in the next major version.
+            |Replace `true` by `all_success`, `false` by `any_success`. Look at documentation for details. 
+        """.trimMargin()
```
Malinskiy: Probably a bit better wording and specify next version (we haven't done a major version so needs to be either an actual version or a minor version):
```
You are using deprecated strict_mode configuration: [true, false].
This API will be removed in 0.7.0
Replace `true` with `all_success`, `false` with `any_success`. See documentation for details. 
```

core/src/main/kotlin/com/malinskiy/marathon/execution/TestBatchResults.kt
Malinskiy: Not really, just an FYI that run stopped can happen if the run was actually stopped and at the same time because it has timed out

Malinskiy: Up to you if you want to create another completion reason for timeout

core/src/main/kotlin/com/malinskiy/marathon/execution/strategy/StrictMode.kt
Malinskiy: Yeah, this sound better

vendor/vendor-android/adam/src/main/kotlin/com/malinskiy/marathon/android/adam/AndroidDeviceTestRunner.kt
Malinskiy: You are right :+1:

vendor/vendor-android/base/src/main/kotlin/com/malinskiy/marathon/android/executor/listeners/AndroidTestRunListener.kt
Malinskiy: I think we shouldn't add more methods to an already complex lifecycle unless required. Otherwise we need to change all our screenshot/log/filesync listeners

Malinskiy (changes requested):

Malinskiy (dismissed):

.gitignore
```diff
@@ -275,4 +275,18 @@ DerivedData
 *.ipa
 *.xcuserstate
 *.profraw
-sample/ios-app/Marathondevices
\ No newline at end of file
+sample/ios-app/Marathondevices
```
Malinskiy, line 278: This file shouldn't be part of the repo: it's specific to each configuration

core/src/main/kotlin/com/malinskiy/marathon/execution/Scheduler.kt
```diff
@@ -40,29 +40,44 @@ class Scheduler(
...
+             */
+            for (child in job.children) {
+                child.join()
+            }
         } catch (e: TimeoutCancellationException) {
-            job.cancelAndJoin()
-            throw NoDevicesException("")
+            throw NoDevicesException("Timeout ${deviceProvider.deviceInitializationTimeoutMillis} ms reached while waiting devices initialization")
+        } finally {
+            close()
```
Malinskiy, line 70: Why we don't need to join the children now? Is that done as part of the close() call?

core/src/main/kotlin/com/malinskiy/marathon/execution/Scheduler.kt
```diff
@@ -109,7 +124,19 @@ class Scheduler(
         logger.debug { "device ${device.serialNumber} associated with poolId ${poolId.name}" }
         pools.computeIfAbsent(poolId) { id ->
             logger.debug { "pool actor ${id.name} is being created" }
-            DevicePoolActor(id, configuration, analytics, shard, progressReporter, track, timer, parent, context, testBundleIdentifier)
+            DevicePoolActor(
+                id,
+                configuration,
+                progressReporter,
+                Channel(),
```
Malinskiy, line 131: Why is this just a constructor here? Looks weird

Malinskiy (changes requested):

Malinskiy:
Already implemented

PR #518, opened 2021-06-07
Add support for ANDROID_SDK_ROOT

mabae:
ANDROID_HOME is deprecated. This PR adds support for ANDROID_SDK_ROOT.

See: https://developer.android.com/studio/command-line/variables

### DIFF
--- a/cli/src/main/kotlin/com/malinskiy/marathon/cli/ApplicationView.kt
+++ b/cli/src/main/kotlin/com/malinskiy/marathon/cli/ApplicationView.kt
@@ -40,7 +40,7 @@ fun main(args: Array<String>): Unit = mainBody(
                 .registerModule(JavaTimeModule())
             val configuration = ConfigFactory(mapper).create(
                 marathonfile = marathonfile,
-                environmentReader = SystemEnvironmentReader { System.getenv(it) }
+                environmentReader = SystemEnvironmentReader()
             )
 
             val application = marathonStartKoin(configuration)
--- a/cli/src/main/kotlin/com/malinskiy/marathon/cli/ApplicationView.kt
+++ b/cli/src/main/kotlin/com/malinskiy/marathon/cli/ApplicationView.kt
@@ -40,7 +40,7 @@ fun main(args: Array<String>): Unit = mainBody(
                 .registerModule(JavaTimeModule())
             val configuration = ConfigFactory(mapper).create(
                 marathonfile = marathonfile,
-                environmentReader = SystemEnvironmentReader()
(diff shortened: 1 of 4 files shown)


cli/src/main/kotlin/com/malinskiy/marathon/cli/ApplicationView.kt
```diff
@@ -40,7 +40,7 @@ fun main(args: Array<String>): Unit = mainBody(
                 .registerModule(JavaTimeModule())
             val configuration = ConfigFactory(mapper).create(
                 marathonfile = marathonfile,
-                environmentReader = SystemEnvironmentReader()
+                environmentReader = SystemEnvironmentReader { System.getenv(it) }
```
Malinskiy: A nice touch in order to test the code

cli/src/main/kotlin/com/malinskiy/marathon/cli/args/environment/SystemEnvironmentReader.kt
```diff
@@ -3,6 +3,13 @@ package com.malinskiy.marathon.cli.args.environment
 import com.malinskiy.marathon.cli.args.EnvironmentConfiguration
 import java.io.File
 
-class SystemEnvironmentReader : EnvironmentReader {
-    override fun read() = EnvironmentConfiguration(System.getenv("ANDROID_HOME")?.let { File(it) })
-}
\ No newline at end of file
+class SystemEnvironmentReader(
+    private val environment: (String) -> String?
```
Malinskiy: Should this have a default value of `{ it -> System.getenv(it)}`?

mabae: You may decide.
My experience: Default values are nice, for example, in data classes. However, in logic classes they may make code and tests harder to understand. Especially in case of a static method call, I wouldn't use it as default.

Malinskiy: Since this is a SystemEnvironmentReader and I'd prefer to have a default value for the lambda here because it will encapsulate at least what the value for the lambda is expected to be

mabae: Done

Malinskiy (approved):

Malinskiy:
Thanks a lot @mabae :+1:

Issue #470, opened 2021-03-16
Dynamic parameters to marathonfile


NoNews:
Hi, thanks for the library!

Could you please clarify if it's possible to pass dynamic parameters to marathon file config somehow?
For instance, for filteringConfiguration. (For one case I'm going to run all tests, for another - only specific ones)

Or the only way for now it's always generate new marathonfile by custom codegen?

Malinskiy:
This was a deliberate decision from the get-go with marathon: all your test runs are represented with a Marathonfile. You can have more than one per repository.
The new configuration #413 might change that since this has been a recurring theme with questions.

Can I ask which parts of the config you want to be specified outside of Marathonfile besides test filtering?

NoNews:
Thank you. 

In my case only test filtering + maybe instrumentation arguments

Malinskiy:
Awesome, thanks for the feedback! I'll make sure this gets into the new configuration's CLI flags

snpefk:
Double test filtering

tagantroy:
We have plans for making cli configuration more flexible.
1) My initial idea was to support field overrides, it will make configuration more flexible.
For example override for influx host may look like this:
`-Dconfig.override.analyticsConfiguration.influx.url="http://localhost:8086"`
But I think making test filtering easy to use with this feature is not trivial. 
2) Second idea was about introducing [starlark](https://github.com/bazelbuild/starlark) support as separate tool to make it very flexible(Starlark is a dialect of Python) and you can easily write script to create configuration based on input parameters. Biggest advantages of starlark: Deterministic evaluation and Hermetic execution. 

It would be nice to hear your thoughts on this too :)

Malinskiy:
A not so obvious feature is that `Marathonfile` supports env substitutions, e.g.:

```yaml
name: "sample-app tests"
outputDir: "${MARATHON_OUTPUT_DIR}"
isCodeCoverageEnabled: true
```

```bash
$ MARATHON_OUTPUT_DIR="build/reports/marathon-custom" marathon
```

abhishekbedi1432:
@Malinskiy  Tried this but failed. 
Command line:
`STRATEGY="device-model" /Users/abhishek.bedi/Downloads/marathon-0.7.3/bin/marathon`

MarathonFile:
`name: "sample-ios-app tests"
outputDir: "derived-data/Marathon"
ignoreFailures: true
testClassRegexes: ["^((?!Abstract).)*Tests$"]
[...]

PR #519, opened 2021-06-08
feat(core): test filtering revamp

Malinskiy:
- add FQTN filter (com.example.MyTest#test1)
- add values list and reading values from file for most of the filters

AnnotationDataFilter is an odd one here and is not currently documented. Left as is without modifications

Using Marathonfile the following will filter the execution to only 2 specific tests:
```yaml
filteringConfiguration:
  allowlist:
    - type: "fully-qualified-test-name"
      values:
        - "com.example.MainActivityTest#testText"
        - "com.example.MainActivityAllureTest#testFailure"
  blocklist:
    - type: "package"
      values:
        - "com.example.somepackage"
        - "com.example.otherpackage"
```

Alternatively, if you need to reuse the same Marathonfile for running different tests:

```yaml
filteringConfiguration:
  allowlist:
    - type: "fully-qualified-test-name"
      file: "path/to/Marathonfilterfile" #Can be any file path, be careful with relative to workdir paths though
  blocklist:
    - type: "package"
      file: "path/to/BlockedPackages"
```

The file should contain FQTN entries on each line, empty lines will be ignored.

```bash
$ cat path/to/Marathonfilterfile
com.example.MainActivityTest#testText
com.example.MainActivityAllureTest#testFailure

$ cat path/to/BlockedPackages
com.example.somepackage
com.example.otherpackage
```

The same pattern will apply to all other filters: instead of specifying regex you can now supply either `values` or a `file` with values

Related to #470


Vacxe (dismissed):

core/src/main/kotlin/com/malinskiy/marathon/execution/filter/SingleValueTestFilter.kt
```diff
@@ -0,0 +1,68 @@
+package com.malinskiy.marathon.execution
+
+import com.malinskiy.marathon.exceptions.ConfigurationException
+import com.malinskiy.marathon.test.Test
+import java.io.File
+
+open class SingleValueTestFilter(
```
Malinskiy, line 7: New type of filter based on a predicate

core/src/main/kotlin/com/malinskiy/marathon/execution/filter/SingleValueTestFilter.kt
```diff
@@ -0,0 +1,68 @@
...
+                null
+            }
+        }
+    }
+
+    private fun readValues(): List<String>? {
+        return values ?: fileValuesCache
+    }
+
+    override fun validate() {
```
Malinskiy, line 27: Validation is needed since only one of parameters is needed

Vacxe (approved):

PR #520, opened 2021-06-09
feat(core): add fragmentation filter

### LINKED ISSUE #353: How to execute tests against only 1 shard of X shards?

Malinskiy:
This is a test filter similar to sharded test execution that AOSP provides https://source.android.com/devices/tech/test_infra/tradefed/architecture/advanced/sharding

It is intended to be used in situations where it is not possible to connect multiple execution devices to marathon, e.g. CI setup that can schedule parallel jobs each containing a single execution device.

This is a dynamic programming technique, hence the results will be sub-optimal compared to connecting multiple devices to the same test run

Apply as following:
```yaml
filteringConfiguration:
  allowlist:
    - type: "fragmentation"
      index: 0
      count: 10
```

If you want to dynamically pass the index of the test run you can use yaml envvar interpolation, e.g.:
```yaml
filteringConfiguration:
  allowlist:
    - type: "fragmentation"
      index: ${MARATHON_FRAGMENT_INDEX}
      count: ${MARATHON_FRAGMENT_COUNT}
```

and then execute the testing as following:

```bash
$ MARATHON_FRAGMENT_COUNT=10 MARATHON_FRAGMENT_INDEX=0 marathon
```

The implementation currently makes a stable decision based on the FQTN. Assuming the count of fragments doesn't change, the test will always be executed in the fragment under the same index. Essentially this makes newly added tests randomly distributed between fragments while guaranteeing that tests are tied to a specific fragment. This execution model doesn't account however for the test duration so an expected edge case is that a particular fragment might exist that will be executed for much longer than other fragments.

Fixes #353

### DIFF
--- a/core/src/main/kotlin/com/malinskiy/marathon/execution/filter/FragmentationFilter.kt
+++ b/core/src/main/kotlin/com/malinskiy/marathon/execution/filter/FragmentationFilter.kt
@@ -1,0 +1,51 @@
+package com.malinskiy.marathon.execution.filter
+
+import com.malinskiy.marathon.exceptions.ConfigurationException
+import com.malinskiy.marathon.execution.TestFilter
+import com.malinskiy.marathon.extension.md5
+import com.malinskiy.marathon.log.MarathonLogging
+import com.malinskiy.marathon.test.Test
+import com.malinskiy.marathon.test.toTestName
+import java.math.BigInteger
+
+/**
+ * This is a test filter similar to sharded test execution that AOSP provides
+ * https://source.android.com/devices/tech/test_infra/tradefed/architecture/advanced/sharding
+ *
+ * It is intended to be used in situations where it is not possible to connect multiple execution devices to marathon, e.g.
+ * CI setup that can schedule parallel jobs each containing a single execution device.
+ *
+ * This is a dynamic programming technique, hence the results will be sub-optimal compared to connecting multiple devices to the same test
+ * run
+ */
+class FragmentationFilter(private val index: Int, private val count: Int) : TestFilter {
+    private val logger = MarathonLogging.logger {}
+    private val power by lazy { BigInteger.valueOf(count.toLong()) }
+    private val remainder by lazy { BigInteger.valueOf(index.toLong()) }
+    private val predicate: (Test) -> Boolean = {
+        /**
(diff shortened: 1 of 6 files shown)


docs/_posts/2018-11-19-configuration.md
```diff
@@ -120,7 +120,8 @@ Here you will find a list of currently supported configuration parameters and ex
 additional parameters might not be supported by all vendor modules. If you find that something doesn't work - please submit an issue for a
 vendor module at fault.
 
-* TOC {:toc}
+* TOC
```
Malinskiy, line 123: Autoformatting removed TOC

core/src/main/kotlin/com/malinskiy/marathon/extension/StringExtensions.kt
```diff
@@ -1,9 +1,20 @@
...
     } else {
         "$prefix.$this"
     }
 }
+
+fun String.md5(): BigInteger {
+    return BigInteger(MD5.md5.digest(toByteArray(charset = Charsets.UTF_8)))
+}
+
+object MD5 {
```
Malinskiy, line 18: Not sure if we want to separate this yet into a different class

core/src/main/kotlin/com/malinskiy/marathon/execution/filter/FragmentationFilter.kt
```diff
@@ -0,0 +1,55 @@
...
+        actualRemainder == remainder
+    }
+
+    override fun validate() {
+        if (index < 0) throw ConfigurationException("Fragment index [$index] should be >= 0")
+        if (count < 0) throw ConfigurationException("Fragment count [$count] should be >= 0")
+        if (index >= count) throw ConfigurationException("Fragment index [$index] should be less than count [$count]")
+
+        logger.info { "Executing test fragment $index out of $count" }
+        logger.warn {
```
Malinskiy, line 46: This warning should encourage people to migrate to different setups

Vacxe (approved):

PR #528, opened 2021-06-16
Cut long class names

Vacxe:


core/src/main/kotlin/com/malinskiy/marathon/io/FileManager.kt
```diff
@@ -65,8 +68,15 @@ class FileManager(private val output: File) {
     private fun createFilename(fileType: FileType, testBatchId: String): String =
         "$testBatchId.${fileType.suffix}"
 
-    private fun createFilename(test: Test, fileType: FileType, testBatchId: String? = null): String =
-        "${test.toTestName()}${testBatchId?.let { "-$it" } ?: ""}.${fileType.suffix}"
+    private fun createFilename(test: Test, fileType: FileType, testBatchId: String? = null): String {
+        val testSuffix = "${testBatchId?.let { "-$it" } ?: ""}.${fileType.suffix}"
+        val testName = test.toTestName().take(256 - testSuffix.length)
```
tagantroy, line 73: What about conflicts? For example: We have two tests with very long names with difference only in last character, after we take `256 - testSuffix.length` they will have equal names. I think in this case it makes more sense to use hash or uuid instead of shortening names or combine both methods together short name + hash/uuid.

Malinskiy: We can either make this human-readable or unique

tagantroy: I would say correct or human-readable :)

Malinskiy: I can see your preference here :) I don't think we can remove the human-readable part though - people sometimes browse the files and they would be completely useless if they're just UUIDs

tagantroy: I think having correct and human-readable filenames is possible, but maybe a bit complicated. Calculating diff between two test names and using it for smart shortening algorithm may work, but sounds too complicated. I think we can go with the existing solution, but also add logs(with error level) in case we have a conflict.

core/src/main/kotlin/com/malinskiy/marathon/io/FileManager.kt
```diff
@@ -65,8 +68,15 @@ class FileManager(private val output: File) {
...
+        val fileName = "$testName$testSuffix"
+        if(testName.length + testSuffix.length > 256) {
+            log.warn("File name length cannot excess 256 characters and has been trimmed to $fileName. This happened because the combination of file path, test class name, and test name is too long.")
```
Malinskiy: Can we change this to error please?

Malinskiy (approved):

PR #539, opened 2021-07-06
On-device test parser

### LINKED ISSUE #393: [Android] Named Parameterized tests will stay uncomplete
**Describe the bug**

When you have tests in a project with parameterized and give them a name, marathon will fail to detect their results and mark them as uncompleted

**To Reproduce**

Change the `ParameterizedTest.kt` to have this as params: `@Parameterized.Parameters(name = "compute test for input {0}: {1}")`

**Expected behavior**

The tests run and succeed

**Logs and reports**

Happens all the time for named parameterized builds

**Devices (please complete the following information):**
 - Device: Any Android device
 - OS: Any Android version

**Additional context**

I have had a look to see if I can open a PR but I failed to see where the best start would have been. Also I was struggling to see if there was a good way to debug things like this?

Malinskiy:
Enabled on adam by default
Currently loses any annotations associated with tests unless a listener is attached to the test parsing, so annotation filter will not work. This is because `am instrument` doesn't output any annotations.

Closes #393
Depends on #538


Malinskiy:
There is an experimental way to extract annotations using on-device parser: https://github.com/Malinskiy/adam/blob/feature/test-meta-listener/android-junit4-listener/src/main/kotlin/com/malinskiy/adam/junit4/android/listener/TestAnnotationProducer.kt

The idea is to attach a RunListener to the execution using `am instrument -e listener TestAnnotationProducer` and parse the metrics outputted by test. One of those metrics will contain a list of annotations.

In order to use this users will have to declare a `androidTestImplementation` dependency with such listener and add an an `instrumentationArgs` parameter to the test parser:

```
vendorConfiguration:
  type: "Android"
....
  vendor: ADAM
  testParserConfiguration:
    type: "remote"
    instrumentationArgs:
      listener: "com.malinskiy.adam.junit4.android.listener.TestAnnotationProducer"
```

Then in the parser annotations will be picked up if present. To allow contract changes (e.g. annotation values) we have to version the instrumentation parameter name.

vendor/vendor-android/adam/src/main/kotlin/com/malinskiy/marathon/android/adam/OnDeviceTestParser.kt
```diff
@@ -0,0 +1,143 @@
...
+
+        val provider = AdamDeviceProvider(configuration, vendorConfiguration, Track(), SystemTimer(Clock.systemDefaultZone()))
+        provider.initialize(vendorConfiguration)
+        val channel = provider.subscribe()
+
+        try {
+            for (update in channel) {
+                if (update is DeviceProvider.DeviceEvent.DeviceConnected) {
+                    val device = update.device as AdamAndroidDevice
+                    return parseTests(device, configuration, vendorConfiguration, testBundles)
```
tagantroy: First connected device will be used for test parsing? Should we have some retry logic here to prevent flakiness if connection is not stable and so on?

Malinskiy: Sure, will do

tagantroy (dismissed):

tagantroy (approved):

PR #571, opened 2021-10-11
Shuffle tests

grab-parry:
Add ability to shuffle test run order. New optional configuration key, shuffleRunOrder, defaults to false if it is not defined in marathon file. To shuffle test run order, include shuffleRunOrder: true in marathon file

### DIFF
--- a/cli/src/main/kotlin/com/malinskiy/marathon/cli/args/FileConfiguration.kt
+++ b/cli/src/main/kotlin/com/malinskiy/marathon/cli/args/FileConfiguration.kt
@@ -35,6 +35,7 @@ data class FileConfiguration(
 
     var testBatchTimeoutMillis: Long?,
     var testOutputTimeoutMillis: Long?,
+    var shuffleRunOrder: Boolean?,
     var debug: Boolean?,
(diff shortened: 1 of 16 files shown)


Malinskiy:
Hey @grab-parry, thanks a lot for the proposed feature!

Have you thought about extending https://github.com/MarathonLabs/marathon/blob/develop/core/src/main/kotlin/com/malinskiy/marathon/execution/strategy/SortingStrategy.kt to achieve the same behaviour?

Logically you're trying to sort the queue and we already have a primitive to do exactly that, so using the proposed flag and using a sorting strategy would be very confusing.

grab-parry:
Hi, I've considered but I'm limited by the SortingStrategy interface which implements a comparator.

Malinskiy:
@grab-parry I believe you can achieve the same behaviour by returning a random comparator that returns a random but deterministic result. WDYT? e.g. -1, 0, 1, 1,0 , -1, etc. This would effectively shuffle the entire queue

Malinskiy:
Not sure if this is still needed/active?

Issue #543, opened 2021-08-03
Issues running under AGP 4.0.1 and 4.2.2


mattinger:
Project Setup:

Gradle 6.7.1
AGP 4.0.1/4.2.2
Kotlin 1.5.20

When i'm trying to run marathon under those conditions, i'm getting the following error

java.lang.NoSuchMethodError: kotlinx.coroutines.channels.ChannelKt.Channel$default(ILkotlinx/coroutines/channels/BufferOverflow;Lkotlin/jvm/functions/Function1;ILjava/lang/Object;)Lkotlinx/coroutines/channels/Channel;
Finish BugSnag

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':app:marathonComcastDebugAndroidTest'.
> Tests failed! See /....

the resulting stack trace and debug logs don't really reveal anything useful.

Caused by: org.gradle.api.GradleException: Tests failed! See .../index.html
        at com.malinskiy.marathon.MarathonRunTask.runMarathon(MarathonRunTask.kt:127)
I suspect it's the NoSuchMethodError that is the issue.

The odd part is that it works fine with AGP 7.0.0 and kotlin 1.5.20

mattinger:
Looking at MarathonRunner, it seems like the error is actually just getting swallowed and a "Tests Failed" message is output at the indicated line above.  I can also tell from the logcat that the app is never even getting deployed to the VM.

mattinger:
Further debugging puts the issue here:

```
declaringClass = "com.malinskiy.adam.transport.vertx.VertxSocketKt"
methodName = "toChannel"
fileName = "VertxSocket.kt"
lineNumber = 214
```

Malinskiy:
Hey @mattinger, thanks for submitting this!

This particular problem you're experiencing is due to the fact that your build controls the classpath of marathon execution and it looks like the version of coroutines in your classpath is different to the one used by marathon. Please read the section on [choosing the distribution](https://marathonlabs.github.io/marathon/doc/downloading.html#choosing-the-distribution) of marathon.

You may be able to get around this error by forcing a particular version of coroutines via buildscript classpath (e.g. https://github.com/MarathonLabs/marathon/blob/develop/sample/android-app/build.gradle.kts#L17), but that still doesn't provide any guarantee whatsoever if your environment didn't upgrade some library that marathon depends on.

trevjonez:
[...]

Issue #460, opened 2021-02-23
Trouble with applying marathon gradle plugin


sundetova:
Hello, I tried to use marathon gradle plugin, but receive next error: 

```
Could not find ch.frankel.log4k:slf4k-api:1.0.0.
Searched in the following locations:
  - https://dl.bintray.com/qameta/maven/ch/frankel/log4k/slf4k-api/1.0.0/slf4k-api-1.0.0.pom
  - https://dl.google.com/dl/android/maven2/ch/frankel/log4k/slf4k-api/1.0.0/slf4k-api-1.0.0.pom
  - https://repo.maven.apache.org/maven2/ch/frankel/log4k/slf4k-api/1.0.0/slf4k-api-1.0.0.pom
Required by:
    project :appName > com.malinskiy.marathon:marathon-gradle-plugin:0.6.1 > com.malinskiy.marathon:core:0.6.1
```

My app build.gradle:
```
plugins {
    id 'com.android.application'
    id 'kotlin-android'
    id 'kotlin-android-extensions'
    id 'marathon' version '0.6.1'
}
```

settings.gradle:
```
pluginManagement {
    repositories {
        maven { url = uri("https://dl.bintray.com/qameta/maven") }
        google()
        mavenCentral()
    }
    resolutionStrategy {
        eachPlugin {
            if (requested.id.id == "marathon") {
                useModule("com.malinskiy.marathon:marathon-gradle-plugin:0.6.1")
            }
        }
    }
}

rootProject.buildFileName = "build.gradle"
include ':appName'
```
I tried to add to my app build.gradle:
`implementation 'ch.frankel.log4k:slf4k-api:1.0.0'
`
But no success. Could you help me?

Malinskiy:
Hey @sundetova! This library is coming from the jCenter. We have not done (as of yet) any review of the dependencies coming from jCenter in regards to their shutdown beginning of the next year.

For now, can you please add the jCenter to the repositories list? If that is not possible, your only option is to switch to CLI as recommended by the docs in the first place.

sundetova:
Thank you for the answer. I added it to settings.gradle repository:
```
pluginManagement {
    repositories {
        mavenCentral()
        maven { url = uri("https://dl.bintray.com/qameta/maven") }
        google()
        jcenter()
    }
    resolutionStrategy {
        ...
}
```
but received next error:
```
Unable to find method 'com.fasterxml.jackson.core.JsonFactory.requiresPropertyOrdering()Z'.
[...]

Issue #458, opened 2021-02-18
0.6.1 -> java.lang.NoSuchMethodError: com.android.ddmlib.AndroidDebugBridge.init(Lcom/android/ddmlib/AdbInitOptions;)V


Georgevik:
Hi

We are using Marathon 0.6.0 in our code base. Since 0.6.1 was released, we can't update to 0.6.1 because when we run the test this error is displayed:

```
java.lang.NoSuchMethodError: com.android.ddmlib.AndroidDebugBridge.init(Lcom/android/ddmlib/AdbInitOptions;)V
```

What are we missing?

Malinskiy:
Hey @Georgevik 

This is most likely classpath manipulation issue: we've updated ddmlib to 27.1.2 in #438 which introduced `AdbInitOptions` class in the first place.

If you're using gradle (and we've had a lot of issues with gradle's classpath before), then your only option is to force the ddmlib version in the `classpath` similar to https://github.com/Malinskiy/marathon/blob/develop/sample/android-app/build.gradle.kts#L18

This is mentioned in the [docs](https://malinskiy.github.io/marathon/doc/downloading.html#choosing-the-distribution) explicitly as a problem with gradle plugin in general. We're working on changes to architect gradle plugin in such a way that it will call to CLI which doesn't have this sort of problems.

Can you please confirm that you're using gradle plugin?

Thanks

asos-jorgegonzalez:
Yes, we are using Android Gradle Plugin 4.0. I will force the ddmlib version using `classpath` and let you know. 

To use Gradle plugin to call CLI sounds great! Gradle sometime (most) is a headache

Georgevik:
Yes, we are using Android Gradle Plugin 4.0. I will force the ddmlib version using classpath and let you know.

To use Gradle plugin to call CLI sounds great! Gradle sometime (most) is a headache

Malinskiy:
@Georgevik any luck forcing the ddmlib version?

Georgevik:
Negative. I tried to force the `ddmlib` version but I couldn't. What I did is to remove Marathon from gradle and run it using CLI. Now it is working fine

Malinskiy:
Closed in 0.7.0 by #573

Issue #447, opened 2021-02-07
java.lang.NoSuchMethodError when Adam enabled in kts


tagantroy:
**Describe the bug**
Gradle plugin is broken. 

**To Reproduce**
1. Update sample app configuration
```
marathon {
    vendor = com.malinskiy.marathon.android.VendorType.ADAM // <-- add this to sample app  
    instrumentationArgs {
        put("debug", "false")
    }
}
```
2. Run `./gradlew marathonDebugAndroidTest`

**Expected behavior**
Run marathon with Adam vendor

**Logs and reports**
```
Output: /home/ivanbalaksha/work/marathon/sample/android-app/app/build/reports/marathon/debugAndroidTest
Ignore failures: false
java.lang.NoSuchMethodError: kotlin.jvm.internal.MutablePropertyReference1Impl.<init>(Ljava/lang/Class;Ljava/lang/String;Ljava/lang/String;I)V
Finish BugSnag
```

**Additional context**
Not 100% sure, but I think it is caused by incompatibility between Kotlin 1.4(used in Marathon) and Kotlin version in gradle kts.

simonpoole:
This "goes away" / is fixed if you use gradle 6.8.

Malinskiy:
For your particular setup using 6.8 solves the problem, however it is not just a compatibility matrix since every build's classpath is unique

Malinskiy:
Closed in 0.7.0 by #573

PR #573, opened 2021-10-13
feat(gradle): lite gradle plugin

### LINKED ISSUE #357: org.koin.core.error.KoinAppAlreadyStartedException: A Koin Application has already been started
I'm seeing the following exception

```
 org.koin.core.error.KoinAppAlreadyStartedException: A Koin Application has already been started
```

to reproduce:

```
./gradlew lib:marathonDebugAndroidTest
# while test is still running, cancel it via `ctrl-c`

# try running again. error will be thrown
./gradlew lib:marathonDebugAndroidTest
```

I believe this happens when I cancel an in-flight test via `ctrl-c` while marathon is running a task like `./gradlew lib:marathonDebugAndroidTest`

Stopping the gradle daemon via `./gradlew --stop` fixes this, and allows me to run `./gradlew lib:marathonDebugAndroidTest` again successfully

Malinskiy:
### TODO:

- [X] Separate configuration into it's own module with minimal dependencies
- [x] Serialize and deserialize YAML for configuration
- [x] Gradle plugin serializes configuration into a temporary YAML file
- [x] Embed marathon binary into plugin and unzip
- [x] Exec marathon from gradle plugin
- [x] Update documentation
- [x] Cache CLI inplugin

Due to the adjustments to deserialization logic, this is a breaking change: for YAML configuration a new notation is used to specifying polymorphic types with a `type` field. For example, this was not the case for analytics configuration which used an object.

Gradle configuration will require some changes as well due to class packet changes.

Closes #357, #442, #447, #458, #460, #543


configuration/src/main/kotlin/com/malinskiy/marathon/config/serialization/yaml/SerializeModule.kt
```diff
@@ -0,0 +1,54 @@
...
+
+class SerializeModule(
+    instantTimeProvider: InstantTimeProvider,
+    environmentReader: EnvironmentReader,
+    marathonfileDir: File,
+    fileListProvider: FileListProvider
+) :
+    SimpleModule() {
+    init {
+//        addDeserializer(AnalyticsConfiguration::class.java, AnalyticsConfigurationDeserializer())
```
tagantroy: Do we use these serializer and deserializer?

configuration/src/main/kotlin/com/malinskiy/marathon/config/serialization/yaml/SerializeModule.kt
```diff
@@ -0,0 +1,54 @@
...
+        )
+        addDeserializer(
+            BatchingStrategyConfiguration.FixedSizeBatchingStrategyConfiguration::class.java,
+            FixedSizeBatchingStrategyConfigurationDeserializer(instantTimeProvider)
+        )
+
+        addSerializer(Regex::class.java, RegexSerializer())
+
+        addDeserializer(File::class.java, FileDeserializer(marathonfileDir))
+//        addSerializer(VendorConfiguration::class.java, VendorConfigurationSerializer())
```
tagantroy: same here

marathon-gradle-plugin/src/main/kotlin/com/malinskiy/marathon/gradle/MarathonRunTask.kt
```diff
@@ -0,0 +1,147 @@
...
+
+    @Internal
+    val testVariant: Property<TestVariant> = objects.property()
+
+    @InputDirectory
+    @PathSensitive(PathSensitivity.NAME_ONLY)
+    val sdk: DirectoryProperty = objects.directoryProperty()
+
+    @Internal
+    val marathonExtension: Property<MarathonExtension> = objects.property()
```
tagantroy, line 50: Will it break caching? If config is changed, task should be executed.

configuration/src/main/kotlin/com/malinskiy/marathon/config/serialization/yaml/SerializeModule.kt
Malinskiy: No, they're removed now. I've used the polymorphic type shared logic from jackson

marathon-gradle-plugin/src/main/kotlin/com/malinskiy/marathon/gradle/MarathonRunTask.kt
Malinskiy: This doesn't change anything compared to the last version - https://github.com/MarathonLabs/marathon/pull/573/files#diff-2ea9389f4257c77da0ff7e8d6ca5038642405a2a0095406caaf3c439e4e40ea6L62

Malinskiy: The task will also be executed every time regardless of the properties since we have `outputs.upToDateWhen { false }`

Vacxe (approved):
Great job mate

PR #578, opened 2021-11-05
feat(marathon-gradle-plugin): rework gradle plugin

Malinskiy:
* Handle clean properly
* Install marathon CLI in root project instead of in every module
* Unzip every time

### DIFF
--- a/marathon-gradle-plugin/src/main/kotlin/com/malinskiy/marathon/gradle/DistributionInstaller.kt
+++ b/marathon-gradle-plugin/src/main/kotlin/com/malinskiy/marathon/gradle/DistributionInstaller.kt
@@ -1,31 +1,13 @@
 package com.malinskiy.marathon.gradle
 
 import org.apache.commons.codec.digest.DigestUtils
-import org.gradle.internal.os.OperatingSystem
 import java.io.File
-import java.nio.file.Files
-import java.nio.file.Paths
-import java.nio.file.attribute.PosixFilePermission
 import java.util.zip.ZipFile
 
 class DistributionInstaller {
-    fun install(marathonBuildDir: File): Pair<File, File> {
+    fun install(marathonBuildDir: File) {
         val marathonZip = copyFromResources(marathonBuildDir)
         unzip(marathonZip, marathonBuildDir)
-        return Pair(marathonBuildDir, getPlatformScript(marathonBuildDir))
-    }
-
-    private fun getPlatformScript(marathonBuildDir: File) = when (OperatingSystem.current()) {
-        OperatingSystem.WINDOWS -> {
-            Paths.get(marathonBuildDir.canonicalPath, "cli", "bin", "marathon.bat").toFile()
-        }
-        else -> {
-            val cliPath = Paths.get(marathonBuildDir.canonicalPath, "cli", "bin", "marathon")
-            cliPath.apply {
-                val permissions = Files.getPosixFilePermissions(this)
-                Files.setPosixFilePermissions(this, permissions + PosixFilePermission.OWNER_EXECUTE)
-            }.toFile()
-        }
     }
 
     private fun copyFromResources(marathonBuildDir: File): File {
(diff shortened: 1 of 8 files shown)


marathon-gradle-plugin/src/main/kotlin/com/malinskiy/marathon/gradle/MarathonRunTask.kt
```diff
@@ -40,10 +41,7 @@ open class MarathonRunTask @Inject constructor(objects: ObjectFactory) : Abstrac
     val flavorName: Property<String> = objects.property()
 
     @Internal
-    val applicationVariant: Property<BaseVariant> = objects.property()
-
-    @Internal
-    val testVariant: Property<TestVariant> = objects.property()
+    val applicationBundles: ListProperty<GradleAndroidTestBundle> = objects.listProperty()
```
Malinskiy, line 48: Instead of only running one module this, in theory, allows us to create a task to run all the modules in parallel.

marathon-gradle-plugin/src/main/kotlin/com/malinskiy/marathon/gradle/MarathonUnpackTask.kt
```diff
@@ -0,0 +1,26 @@
+package com.malinskiy.marathon.gradle
+
+import org.gradle.api.DefaultTask
+import org.gradle.api.model.ObjectFactory
+import org.gradle.api.tasks.TaskAction
+import java.io.File
+import javax.inject.Inject
+
+open class MarathonUnpackTask @Inject constructor(objects: ObjectFactory) : DefaultTask() {
```
tagantroy: Objects field is unused, right?

marathon-gradle-plugin/src/main/kotlin/com/malinskiy/marathon/gradle/MarathonPlugin.kt
```diff
@@ -11,20 +11,31 @@ import org.gradle.api.Plugin
...
 class MarathonPlugin : Plugin<Project> {
 
     override fun apply(project: Project) {
         val logger = project.logger
         logger.info("Applying marathon plugin")
-        val extension: MarathonExtension = project.extensions.create("marathon", MarathonExtension::class.java, project)
+        project.extensions.create("marathon", MarathonExtension::class.java, project)
 
+        val rootProject = project.rootProject
+        synchronized(rootProject) {
```
tagantroy: Do we need synchronized here?

marathon-gradle-plugin/src/main/kotlin/com/malinskiy/marathon/gradle/DistributionInstaller.kt
```diff
@@ -9,16 +9,15 @@ import java.nio.file.attribute.PosixFilePermission
 import java.util.zip.ZipFile
 
 class DistributionInstaller {
-    fun install(marathonBuildDir: File): File {
+    fun install(marathonBuildDir: File): Pair<File, File> {
```
tagantroy: Looks like we never use result of this function

tagantroy (dismissed):

Vacxe (dismissed):

marathon-gradle-plugin/src/main/kotlin/com/malinskiy/marathon/gradle/DistributionInstaller.kt
Malinskiy: You're correct, but this is mostly because I don't know how to propagate these between tasks except just with a convention

marathon-gradle-plugin/src/main/kotlin/com/malinskiy/marathon/gradle/MarathonPlugin.kt
Malinskiy: Do you know if gradle can use `apply` block on multiple projects at the same time? This sync is to have only one root project configuration basically

tagantroy (approved):

PR #580, opened 2021-11-09
Add groovy sample for file sync configuration

alirezaafkar:


Malinskiy:
0.7.0 will introduce a breaking change: the configuration classes will change packages due to the complete rework of the gradle plugin. for this case I suggest not using FQDN or using the new `com.malinskiy.marathon.config.vendor.android` package

docs/_posts/2018-11-19-android.md
```diff
@@ -512,6 +512,41 @@ vendorConfiguration:
...
 ```
 
+{% endtab %} {% tab file-sync Gradle %}
+
+```kotlin
+marathon {
+    marathon {
+    fileSyncConfiguration {
+      pull.add(
+        new com.malinskiy.marathon.android.configuration.FileSyncEntry(
```
Malinskiy, line 522: FQDN should be simplified to just the simple class name or new package `com.malinskiy.marathon.config.vendor.android` should be used that will be effective as of 0.7.0

PR #588, opened 2021-11-15
feat(android): no external storage permissions required

### LINKED ISSUE #570: Use app writeable directory for coverage files on Android 11
**Is your feature request related to a problem? Please describe.**
Currently generating coverage files on Android 11 and later requires requesting and granting the global MANAGE_EXTERNAL_STORAGE permission to the app under testing, as the /sdcard/coverage directory created is otherwise not accessible by the app. While this workaround can be restricted to the debug version of the app, this is naturally sub-optimal from a testing perspective as the app is running with permissions that it will not have in production and which will hide any file access related issues.

**Describe the solution you'd like**
Configure the instrumentation runner to write to a directory that is writable by the app under testing, for example /sdcard/Download/coverage.

Alternatively the coverage file destination directory could be made configurable, but IMHO that is not really worth the effort. 

**Describe alternatives you've considered**
The alternative is granting the permission, which has negative side effects as outlined above.

**Additional context**

A quick test that the suggested destination actually works can be done with (groovy gradle, MANAGE_EXTERNAL_STORAGE not requested)

```
...
    instrumentationArgs { 
        put("coverageFile", "/sdcard/Download/coverage/coverage.ec")
    }
    fileSyncConfiguration {
        pull.add(
            new com.malinskiy.marathon.android.configuration.FileSyncEntry(
                "Download/coverage",
                com.malinskiy.marathon.android.configuration.AggregationMode.DEVICE_AND_POOL
            )
        )
   }
...
````

### PR DESCRIPTION
**Pull request by Malinskiy**
Currently, coverage and other test artifacts are assumed to always be stored somewhere in the external storage.

This PR adds support for using internal app storage (e.g. using [Context#getFilesDir()](https://developer.android.com/reference/android/content/Context#getFilesDir()) for storing code coverage, allure-results and more.

This allows users of marathon to remove the permissions such as:
```xml
  <uses-permission android:name="android.permission.MANAGE_EXTERNAL_STORAGE"/>
  <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />
  <uses-permission android:name="android.permission.READ_EXTERNAL_STORAGE" />
```
from AndroidManifest.xml if the only reason for them is testing.
This in turn brings production and test versions of the codebase more aligned and helps catch external storage-related bugs on test versions of the app rather than in production.

Depends on changes in allure https://github.com/allure-framework/allure-kotlin/pull/65 to produce allure-results in `/data/data/$pkg/files/allure-results`

Fixes #570


sample/android-app/buildSrc/src/main/kotlin/Versions.kt
```diff
@@ -12,7 +12,7 @@ object Versions {
     val junit = "4.13.2"
     val appCompat = "1.2.0"
     val constraintLayout = "2.0.4"
-    val allure = "2.1.2"
+    val allure = "1.0-SNAPSHOT"
```
Malinskiy: Should be updated if allure-kotlin accepts PR

vendor/vendor-android/base/src/main/kotlin/com/malinskiy/marathon/android/BaseAndroidDevice.kt
```diff
@@ -100,17 +101,19 @@ abstract class BaseAndroidDevice(
         if (androidConfiguration.allureConfiguration.enabled) {
             androidConfiguration.fileSyncConfiguration.pull.add(
                 FileSyncEntry(
-                    androidConfiguration.allureConfiguration.relativeResultsDirectory,
-                    AggregationMode.TEST_RUN
+                    relativePath = androidConfiguration.allureConfiguration.relativeResultsDirectory,
+                    pathRoot = PathRoot.APP_DATA,
```
Malinskiy: Should be changed back to external storage if allure-kotlin doesn't accept the PR

Malinskiy:
Will not pass CI until allure-kotlin decision is made

PR #611, opened 2022-02-25
Add integration with Teamcity status message

Vacxe:


tagantroy:
Please provide more details about this feature, also we will need to update docs after you merge it.

integrations/buildsystem/build.gradle.kts
```diff
@@ -0,0 +1,20 @@
+import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
+
+plugins {
+    `java-library`
+    id("org.jetbrains.kotlin.jvm")
+    id("org.jetbrains.dokka")
+}
+
+Deployment.initialize(project)
```
Malinskiy: If you're adding a new module - it should be published as well. I don't think we need to create a separate jar for this though

configuration/src/main/kotlin/com/malinskiy/marathon/config/Configuration.kt
```diff
@@ -73,7 +75,8 @@ data class Configuration private constructor(
             "debug" to debug.toString(),
             "screenRecordingPolicy" to screenRecordingPolicy.toString(),
             "vendorConfiguration" to vendorConfiguration.toString(),
-            "deviceInitializationTimeoutMillis" to deviceInitializationTimeoutMillis.toString()
+            "deviceInitializationTimeoutMillis" to deviceInitializationTimeoutMillis.toString(),
+            "buildSystem" to buildSystem
```
Malinskiy: This probably should be a separate configuration class, e.g. BuildSystemConfiguration -> TeamCityConfiguration instead of just an enum. Otherwise we don't make it extensible in the future with parameters to the build system (e.g. ignore X, print more, etc)

Malinskiy (changes requested):

Vacxe:
See: #964

PR #617, opened 2022-03-17
Use fully qualified test name in allure report

Litemn:
io.qameta.allure:allure-kotlin-model use FQN for fullName and pkg.clazz for suite


core/src/main/kotlin/com/malinskiy/marathon/report/allure/AllureReporter.kt
```diff
@@ -98,7 +98,7 @@ class AllureReporter(val configuration: Configuration, private val outputDirecto
                 mutableListOf(
                     ResultsUtils.createHostLabel().setValue(device.serialNumber),
                     ResultsUtils.createPackageLabel(test.pkg),
-                    ResultsUtils.createTestClassLabel(test.clazz),
+                    ResultsUtils.createTestClassLabel(suite),
```
Malinskiy, line 101: Can you please provide some docs on why test class should be specified using fqcn here and not just a simple class name?
Same for the full name above, would like to understand where is the source of truth for these

Malinskiy (changes requested):

core/src/main/kotlin/com/malinskiy/marathon/report/allure/AllureReporter.kt
Malinskiy: @Litemn ping

Malinskiy: @Litemn another ping

Litemn: Here some example of test result generated by allure lib
[allure kotlin](https://github.com/allure-framework/allure-kotlin/blob/master/allure-kotlin-junit4/src/main/kotlin/io/qameta/allure/kotlin/junit4/AllureJunit4.kt#L139)
[allure java](https://github.com/allure-framework/allure-java/blob/b782b29f45e1720481f2c853eb9a03cd13bc11cc/allure-junit-platform/src/main/java/io/qameta/allure/junitplatform/AllureJunitPlatformUtils.java#L39)
Screenshot 2022-04-22 at 15 04 33
example-junit-5-allure

kotlin and java implemantation work same and use FQN for fullName and package for suite

Malinskiy (approved):

PR #630, opened 2022-04-17
Install extra applications if they specified in Marathonfile

kavabunga6:
Pull request for https://github.com/MarathonLabs/marathon/issues/629


Malinskiy:
Hey @kavabunga6, thanks a lot for the PR!
Could you please describe the use-case for this feature? I’d like to know when you would want something like this
Thanks!

kavabunga6:
> Hey @kavabunga6, thanks a lot for the PR! Could you please describe the use-case for this feature? I’d like to know when you would want something like this Thanks!

Use-case very simple:
Our tests are in a separate reposity and we should pre-install on device main application
So, we need use 3 apk: main-app.apk, app-debug.apk, app-debug-androidTest.apk

И по-русски :)
Наши тесты лежат в отдельном репозитории и требуют предустановку основного приложения. Соответственно у нас на входе 3 apk: main-app.apk, app-debug.apk, app-debug-androidTest.apk

trevjonez:
This would also be useful for installing things such as the android a11y suite before tests start on an ephemeral emulator. https://play.google.com/store/apps/details?id=com.google.android.marvin.talkback

kavabunga6:
> This would also be useful for installing things such as the android a11y suite before tests start on an ephemeral emulator. https://play.google.com/store/apps/details?id=com.google.android.marvin.talkback

This solution implies extra scripts or actions in preparing the test
Both extra points of failure.
In the second case, this can also lead to extra repetitions of the test

It is better when one point is responsible for the entire flow of the same type of actions

P.S. sorry for my english :)

Malinskiy:
@kavabunga6 LGTM. Only thing is this might benefit from some tests

kavabunga6:
> @kavabunga6 LGTM. Only thing is this might benefit from some tests

@Malinskiy done, added two tests

Malinskiy (approved):

Malinskiy:
@kavabunga6 merged, thanks for your contribution! :tada:

Issue #620, opened 2022-03-21
Add Support for TestPlan


SaketGupte:
**Is your feature request related to a problem? Please describe.**
iOS introduced feature called TestPlan back in 2020. It allows for running Unit and UI Test multiple times with different configurations.
Configurations can be different language and region or Simulated location. Think of LTR and RTL snapshot tests.
Without TestPlan each configuration would require scheme. With TestPlan you can have all the configurations run together.

**Describe the solution you'd like**
Marathon should support TestPlan as well. 

**Additional context**
A more detailed explaination of what a TestPlan is can be found here: https://useyourloaf.com/blog/xcode-test-plans/

Malinskiy:
Marathon has it's own test filtering and scheduling system. I don't really understand why there is a need to use custom formats that are not even well documented: xctestrun file is only documented in manpages and even then the docs do not match the implementation.

If the source of truth for test list is xctestrun for users then an additional utility is needed that transforms testplan into a simple line-by-line list of tests that can be recognized by marathon https://docs.marathonlabs.io/configuration/filtering#values-file-filtering

PR #686, opened 2022-09-16
[iOS] Added Support for XCResult

abhishekbedi1432:
### Motivation
Marathon for iOS produces an amazing HTML report (in the `outputDir`) but sometimes we need the final `xcresult` bundle for many reasons (say integrating with the CI, uploading the result in repository, etc).

The `xcresult` bundles  are stored at transient paths under Xcode's `derived-data` folder. 

In this PR, we have added an enhancement of providing `xcResultBundlePath` parameter in the _Marathon_ file (under iOS Vendor Config)

As a positive side effect, this helps to avoid flooding Xcode's derived data directory.


README.md
```diff
@@ -6,13 +6,49 @@
...
 ## Main focus
-- **stability** of test execution adjusting for flakiness in the environment and in the tests. 
+- **stability** of test execution adjusting for flakiness in the environment and in the tests.
 - **performance** using high parallelization (handling dozens of devices)
 
 ## Documentation
 
 Please check the official [documentation](https://marathonlabs.github.io/marathon/) for installation, configuration and more
 
+## [iOS Only] Added Support to consume tags for test functions under XCTestCase Subclasses
```
Malinskiy, line 16: Please move the docs info to the https://github.com/MarathonLabs/marathon/blob/develop/docs/_posts/2018-11-19-ios.md. They're automatically published at https://marathonlabs.github.io/marathon/

configuration/src/main/kotlin/com/malinskiy/marathon/config/serialization/ConfigurationFactory.kt
```diff
@@ -55,11 +55,23 @@ class ConfigurationFactory(
                 is VendorConfiguration.IOSConfiguration -> {
                     // Any relative path specified in Marathonfile should be resolved against the directory Marathonfile is in
                     val resolvedDerivedDataDir = marathonfileDir.resolve(configuration.vendorConfiguration.derivedDataDir)
-                    val finalXCTestRunPath = configuration.vendorConfiguration.xctestrunPath?.resolveAgainst(marathonfileDir)
-                        ?: fileListProvider
+                    val resolvedResultBundlePath = marathonfileDir.resolve(configuration.vendorConfiguration.xcResultBundlePath)
+
+                    // Adding support for Test Plan
```
Malinskiy, line 60: Please remove the comment. It doesn't make sense after we merge this.

configuration/src/main/kotlin/com/malinskiy/marathon/config/serialization/ConfigurationFactory.kt
```diff
@@ -55,11 +55,23 @@ class ConfigurationFactory(
...
+                    val testPlanName = configuration.vendorConfiguration.xcTestPlan
+
+                    var finalXCTestRunPath = if(!testPlanName.isNullOrEmpty()) {
+                        fileListProvider
                             .fileList(resolvedDerivedDataDir)
-                            .firstOrNull { it.extension == "xctestrun" }
-                        ?: throw ConfigurationException("Unable to find an xctestrun file in derived data folder")
+                            .firstOrNull { it.extension == "xctestrun" && it.name.contains("$testPlanName") } ?: throw ConfigurationException("Unable to find matching TestPlan. Please recheck if testplan is enabled")
```
Malinskiy: You can skip creating an additional string here:
`it.name.contains("$testPlanName")` -> `it.name.contains(testPlanName)`
It might also be a good idea to use startsWith or equals instead of contains. I'm not sure if there is a contract on the file name here

core/src/main/kotlin/com/malinskiy/marathon/Marathon.kt
```diff
@@ -79,6 +80,10 @@ class Marathon(
         val tests = applyTestFilters(parsedTests)
         val shard = prepareTestShard(tests, analytics)
 
+        log.info("\n\n\n **** Marathon File Params **** \n")
```
Malinskiy, line 83: This is a no-no: configuration might contain sensitive credentials, and we can't print them to the stdout

Malinskiy: If you really want to do this then we need to implement some kind of masking for sensitive fields

core/src/main/kotlin/com/malinskiy/marathon/test/Test.kt
```diff
@@ -6,7 +6,8 @@ data class Test(
     val pkg: String,
     val clazz: String,
     val method: String,
-    val metaProperties: Collection<MetaProperty>
+    val metaProperties: Collection<MetaProperty>,
+    val tags: Collection<String> = emptyList()
```
Malinskiy, line 10: MetaProprety is an abstraction over tags, annotations and so on. We should just reuse MetaProperty here

core/src/main/kotlin/com/malinskiy/marathon/Marathon.kt
```diff
@@ -146,7 +151,9 @@ class Marathon(
         }
         configuration.filteringConfiguration.allowlist.forEach { tests = it.toTestFilter().filter(tests) }
         configuration.filteringConfiguration.blocklist.forEach { tests = it.toTestFilter().filterNot(tests) }
-        return tests
+        val iosConfig = configuration.vendorConfiguration as? VendorConfiguration.IOSConfiguration
+
+        return tests.filter { it.tags.contains(iosConfig?.xcTestRunnerTag) }
```
Malinskiy, line 156: This should just be an annotation filter, no vendor-specific logic in the core

sample/ios-app/sample-appUITests/MoreTests.swift
```diff
@@ -27,6 +27,7 @@ class MoreTests: XCTestCase {
         super.tearDown()
     }
 
+    // @Flowers
```
Malinskiy, line 30: Instead of modifying existing tests we should just add more tests with these tags

vendor/vendor-ios/src/main/kotlin/com/malinskiy/marathon/ios/IOSDevice.kt
```diff
@@ -382,3 +390,11 @@ private fun String.toInetAddressOrNull(): InetAddress? {
 
 private fun TestBatch.toXcodebuildArguments(): String =
     tests.joinToString(separator = " ") { "-only-testing:\"${it.pkg}/${it.clazz}/${it.method}\"" }
+
+private fun getTimezone(): String {
```
Malinskiy, line 394: This method sounds like getTimestamp rather than getTimezone

vendor/vendor-ios/src/main/kotlin/com/malinskiy/marathon/ios/IOSTestParser.kt
```diff
@@ -20,6 +21,11 @@ class IOSTestParser(private val vendorConfiguration: VendorConfiguration.IOSConf
      *  marked as skipped in `xctestrun` file.
      */
     override suspend fun extract(): List<Test> {
+
+        if (vendorConfiguration.xcTestRunnerTag.isNullOrEmpty()) {
+            logger.warn { "[iOS] Did not find `xcTestRunnerTag` in iOS Vendor Config. This will result in running all tests" }
```
Malinskiy, line 26: The tag filtering doesn't have anything to do with the responsibility of the parser, this probably doesn't belong here

vendor/vendor-ios/src/main/kotlin/com/malinskiy/marathon/ios/xctestrun/Xctestrun.kt
```diff
@@ -24,41 +24,53 @@ class Xctestrun(inputStream: InputStream) {
...
-
+/*
     private val target = PropertyListKey.TargetName(
         propertyList.keys.firstOrNull { it != PropertyListKey.__xctestrun_metadata__.toKeyString() }
             ?: throw IllegalArgumentException("xctestrun file does not define any testable targets")
     )
+*/
+
+    private val targetMap: Map<String, *> = (
+        ((((((propertyList.valueForKeypath("TestConfigurations") as? Array<Any>)?.first()) as Map<*, *>)
```
Malinskiy, line 35: first() will throw NullPointerException here when `TestConfigurations` is missing. Reproduced on the old sample

Malinskiy (changes requested):
Logic wise we're missing pulling back the xcresult back to the marathon's side

Malinskiy:
Closes #620

abhishekbedi1432:
Pls refer: https://github.com/MarathonLabs/marathon/pull/700 for the feature (Adding support for XCResult). I will close this PR addressing all the comments in https://github.com/MarathonLabs/marathon/pull/700

core/src/main/kotlin/com/malinskiy/marathon/test/Test.kt
abhishekbedi1432: Refer: 
[iOS]Added Support for Cucumberish Tags / Function Annotations https://github.com/MarathonLabs/marathon/pull/702

PR #700, opened 2022-10-26
[iOS]Added support for resultBundlePath

abhishekbedi1432:
### Motivation
Marathon for iOS produces an amazing HTML report (in the outputDir) but sometimes we need the final xcresult bundle for many reasons (say integrating with the CI, uploading the result in repository, etc).

The xcresult bundles are stored at transient paths under Xcode's derived-data folder.

In this PR, we have added an enhancement of providing xcResultBundlePath parameter in the Marathon file (under iOS Vendor Config)

As a positive side effect, this helps to avoid flooding Xcode's derived data directory.


Malinskiy:
Closed in favour of #719

Issue #589, opened 2021-11-15
Feature Request: Standalone marathon yml file generation task


trevjonez:
**Is your feature request related to a problem? Please describe.**
Currently our CI setup generates a marathon file via a bash heredoc setup, that config is then packaged with apk's and sent to an external system for execution.

Being able to utilize the gradle DSL to replace the bash scripting could be a big win for developer UX when running locally as the marathon file would become dynamic and on demand while also letting me capture the configuration in more of a "single source of truth" model.

**Describe the solution you'd like**
Split the [MarathonRunTask](https://github.com/MarathonLabs/marathon/blob/develop/marathon-gradle-plugin/src/main/kotlin/com/malinskiy/marathon/gradle/MarathonRunTask.kt) into a generate config task and a run task. Looking at the existing task logic it is likely there would need to be a few things modified in order to do adhoc generate config tasks, but the existing logic is 99% of the way there, just needs to be decoupled a bit more.

**Describe alternatives you've considered**
As mentioned a heredoc pattern is being used in bash scripts which is fine for CI, but really leaves any local marathon runs as a fairly tedious process for developers that are new to the tool.

Anything I can do to help people avoid manually writing YML I'd like to pursue :laughing:

Malinskiy:
Closed by #701

PR #701, opened 2022-10-30
feature(gradle-plugin): generate Marathonfile for variant task

Malinskiy:
All tasks have as lazy as possible configuration while keeping the main configuration structure sane. We don't want to define every single property in the `Configuration` as a separate `Provider` in each task. To achieve this json serialisation into string is used to pass the whole configuration instance around the tasks.

This change introduces several new tasks split from the previous logic, namely:
* marathonWrapperExtract is the task of copying CLI distribution from resources to temp dir
* marathonWrapper now depends on marathonWrapperExtract and extracts zip into usable distro
* xxxxGenerateMarathonfile will now generate the Marathonfile for a particular flavor and provide marathonfile as `@OutputFile`. This doesn't require any unpacking of the CLI above
* marathonXXX is the old task name to trigger the tests for a particular variant

```
> Task :marathonWrapperExtract UP-TO-DATE
> Task :marathonWrapper UP-TO-DATE
> Task :app:marathonDebugAndroidTestGenerateMarathonfile UP-TO-DATE
```


Malinskiy:
Closes #589

Vacxe (approved):

configuration/src/main/kotlin/com/malinskiy/marathon/config/Configuration.kt
```diff
@@ -145,43 +145,40 @@ data class Configuration private constructor(
...
+        var retryStrategy: RetryStrategyConfiguration = RetryStrategyConfiguration.NoRetryStrategyConfiguration,
+        var filteringConfiguration: FilteringConfiguration = FilteringConfiguration(emptyList(), emptyList()),
+
+        var ignoreFailures: Boolean = false,
+        var isCodeCoverageEnabled: Boolean = false,
+        var fallbackToScreenshots: Boolean = false,
+        var strictMode: Boolean = false,
+        var uncompletedTestRetryQuota: Int = Integer.MAX_VALUE,
+
+        var testClassRegexes: Collection<Regex> = listOf(Regex("^((?!Abstract).)*Test[s]*$")),
```
sergioboi, line 166: Do we have a test about this default regex?

configuration/src/main/kotlin/com/malinskiy/marathon/config/Configuration.kt
```diff
@@ -145,43 +145,40 @@ data class Configuration private constructor(
...
+        var excludeSerialRegexes: Collection<Regex> = emptyList(),
+
+        var testBatchTimeoutMillis: Long = DEFAULT_BATCH_EXECUTION_TIMEOUT_MILLIS,
+        var testOutputTimeoutMillis: Long = DEFAULT_OUTPUT_TIMEOUT_MILLIS,
+        var debug: Boolean = true,
+
+        var screenRecordingPolicy: ScreenRecordingPolicy = ScreenRecordingPolicy.ON_FAILURE,
+
+        var analyticsTracking: Boolean = false,
+        var deviceInitializationTimeoutMillis: Long = DEFAULT_DEVICE_INITIALIZATION_TIMEOUT_MILLIS,
```
sergioboi, line 177: Could we use `Duration` instead of `Long`?

marathon-gradle-plugin/src/main/kotlin/com/malinskiy/marathon/gradle/GradleAndroidTestBundle.kt
```diff
@@ -3,10 +3,15 @@ package com.malinskiy.marathon.gradle
...
+import org.gradle.api.tasks.Nested
+import java.io.Serializable
 
 data class GradleAndroidTestBundle(
-    val apkFolder: DirectoryProperty? = null,
-    val artifactLoader: Property<BuiltArtifactsLoader>? = null,
-    val testApkFolder: DirectoryProperty,
-    val testArtifactLoader: Property<BuiltArtifactsLoader>,
-)
+    @InputDirectory val apkFolder: DirectoryProperty? = null,
```
sergioboi, line 13: It's not a Gradle way. Properties have get() that could return null.
So we could use not nullable Properties without value as default

marathon-gradle-plugin/src/main/kotlin/com/malinskiy/marathon/gradle/task/GenerateMarathonfileTask.kt
```diff
@@ -0,0 +1,116 @@
...
+import org.gradle.api.tasks.OutputFile
+import org.gradle.api.tasks.PathSensitive
+import org.gradle.api.tasks.PathSensitivity
+import org.gradle.api.tasks.TaskAction
+import org.gradle.kotlin.dsl.listProperty
+import org.gradle.kotlin.dsl.property
+import java.io.File
+import javax.inject.Inject
+
+open class GenerateMarathonfileTask @Inject constructor(objects: ObjectFactory) : DefaultTask() {
```
sergioboi, line 30: it could be `abstract class`

marathon-gradle-plugin/src/main/kotlin/com/malinskiy/marathon/gradle/task/GenerateMarathonfileTask.kt
```diff
@@ -0,0 +1,116 @@
...
+    init {
+        group = Const.GROUP
+    }
+
+    @Input
+    val flavorName: Property<String> = objects.property()
```
sergioboi, line 36: fyi
you could use `abstract val flavorName: Property<String>` and Gradle will initialize them for you

marathon-gradle-plugin/src/main/kotlin/com/malinskiy/marathon/gradle/task/GenerateMarathonfileTask.kt
```diff
@@ -0,0 +1,116 @@
...
+
+    @Internal
+    val jsonService: Property<JsonService> = objects.property()
+    
+    @TaskAction
+    fun write() {
+        val json = jsonService.get()
+        
+        // Override stuff coming from AGP
+        val androidConfiguration = vendorConfigurationBuilder.let { json.parseVendor(it.get()) }.apply {
```
sergioboi, line 63: Want to rewrite that place without let trying use Gradle apis.
If you want I could do that

marathon-gradle-plugin/src/main/kotlin/com/malinskiy/marathon/gradle/task/MarathonDownloadTask.kt
```diff
@@ -0,0 +1,21 @@
...
+import com.malinskiy.marathon.gradle.Const
+import org.gradle.api.file.RegularFileProperty
+import org.gradle.api.model.ObjectFactory
+import org.gradle.api.tasks.Copy
+import org.gradle.api.tasks.OutputFile
+import javax.inject.Inject
+
+open class MarathonDownloadTask @Inject constructor(objects: ObjectFactory) : Copy() {
+    init {
+        group = Const.GROUP
```
sergioboi, line 12: It would be great to add `description`

marathon-gradle-plugin/src/main/kotlin/com/malinskiy/marathon/gradle/task/GenerateMarathonfileTask.kt
Malinskiy: or open. As long as gradle can modify the class. If I want to initialize this task from tests - I can. If I use abstract - it's not possible.

marathon-gradle-plugin/src/main/kotlin/com/malinskiy/marathon/gradle/task/GenerateMarathonfileTask.kt
Malinskiy: What do you mean using Gradle APIs? This is just get string, parse it from json. let is used just to map a value in one line

marathon-gradle-plugin/src/main/kotlin/com/malinskiy/marathon/gradle/task/MarathonDownloadTask.kt
Malinskiy: Good catch

marathon-gradle-plugin/src/main/kotlin/com/malinskiy/marathon/gradle/task/GenerateMarathonfileTask.kt
Malinskiy: I don't like abstract code. We write code that actually is executed and can be tested via regular tests.

marathon-gradle-plugin/src/main/kotlin/com/malinskiy/marathon/gradle/GradleAndroidTestBundle.kt
Malinskiy: This is not Gradle. It's a data class that passes wiring for Gradle stuff. Perfectly reasonable to have a nullable apk since libraries do not have any apk, only the test apk.

configuration/src/main/kotlin/com/malinskiy/marathon/config/Configuration.kt
Malinskiy: We can. Not sure if it will break existing configurations. If we want to migrate millis then it probably should be for every field that related to duration here

configuration/src/main/kotlin/com/malinskiy/marathon/config/Configuration.kt
Malinskiy: Parsing of the regex - yes. Running it - scenario tests cover this fake running and verify this

PR #702, opened 2022-10-30
[iOS]Added Support for Cucumberish Tags / Function Annotations

abhishekbedi1432:
### Background
For Android, marathon already supports [Cucumber](https://cucumber.io/docs/cucumber/api/?lang=java) framework for BDD (Behaviour Driven Development) oriented UI Tests but the same is not present for iOS (which uses a similar [Cucumberish](https://github.com/Ahmed-Ali/Cucumberish)). 

### Problem
In order to leverage marathon to support cucumberish (for iOS) we may require a lot of effort.

### Solution
In Cucumber/Cucumberish, we annotate the scenarios/features with tags like `@SmokeTest @Regression`, etc.

Now considering if your project supports Cucumberish or not , there are two approaches:

1. If your iOS project is NOT using Cucumberish framework for writing tests, you can simply annotate your test functions and same will work with marathon.
2. However, if you are using Cucumberish, and trying to move out of it for a more robust solution, you can use [XCTest-Gherkin framework](https://github.com/net-a-porter-mobile/XCTest-Gherkin) to convert your Cucumberish Tests to native UI Tests and then tag as described below.

### How Test Functions Look Like
image

### Please Note
Test Functions can have multiple annotations but marathon will run only those tests which match the `xcTestRunnerTag` i.e. _only one tag at a time_

### Next Steps
We will add functionality to support multiple tags as well in the future :)

### DIFF
--- a/configuration/src/main/kotlin/com/malinskiy/marathon/config/vendor/VendorConfiguration.kt
+++ b/configuration/src/main/kotlin/com/malinskiy/marathon/config/vendor/VendorConfiguration.kt
@@ -83,7 +83,8 @@ sealed class VendorConfiguration {
         @JsonProperty("compactOutput") val compactOutput: Boolean = false,
         @JsonProperty("keepAliveIntervalMillis") val keepAliveIntervalMillis: Long = 0L,
         @JsonProperty("devices") val devicesFile: File? = null,
-    ) : VendorConfiguration() {
+        @JsonProperty("xcTestRunnerTag") val xcTestRunnerTag: String? = null,
+        ) : VendorConfiguration() {
         /**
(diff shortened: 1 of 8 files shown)


Malinskiy:
With the latest change coming in 0.8.0 marathon doesn't have access to source code of the test anymore to align more with other platforms.
I see 2 solutions here:
- Implement some way of discovering tags in runtime (not sure how since Apple ecosystem doesn't even have proper test listing ATM)
- Implement this as a separate tool/command to create a test filtering configuration for Marathonfile. Marathon supports reading test list from a separate file in a format:
```
com.example.MyTest#method1
com.example.MyTest2#method2
...
```

Best way forward would be option 1, but it requires Apple's support

PR #724, opened 2022-12-16
feat(cli): parse subcommand

matzuk:


cli/src/main/kotlin/com/malinskiy/marathon/cli/args/CliCommands.kt
```diff
@@ -0,0 +1,86 @@
...
+}
+
+class Parse(
+    private val starter: (CliConfiguration) -> Unit
+): CliktCommand(name = "parse", help="Print the list of tests without executing them") {
+
+    private val marathonfile by option("--marathonfile", "-m", help="Marathonfile file path")
+        .file()
+        .default(File("Marathonfile"))
+    private val parseOutputFile by option("--output", "-o", help="Output for the result of Parse command execution")
```
Malinskiy: Add an option for format with only `yaml` for now?

matzuk: @Malinskiy  I've added few words that the output format is yaml. Do we need to support other formats?

Malinskiy (approved):

PR #726, opened 2022-12-28
feat(ios): completely rework ios

Malinskiy:
This PR reworks iOS implementation of testing. Short version of changes:
- Add support for local test runs without SSH
- Support XCUITests as well as XCTests
- Support creating, booting, erasing, shutting down simulators with required configurations
- Remove dependency on source code parsing for listing tests
- Require providing only `.app` and `.xctest` (including as zip and ipa and derived data dir)
- Enhance auth support for ssh with password auth and known hosts in OpenSSH server format
- Support specifying XC attachment lifetime
- Support screen recording (video and screenshots assembled into a gif)
- Support lifecycle hooks for each simulator to erase on start or terminate on dispose at the end of the run
- Support granting permissions
- Explicit timeouts for every operation

and more

Docs are updated in this PR, refer to them for more info, until 0.8.0 they will be at [docs.marathonlabs.io/next/ios](https://docs.marathonlabs.io/next/ios)


Nader822:
I'm

PR #753, opened 2023-02-09
fix(marathon-gradle-plugin): fixed groovy filtering configuration

### LINKED ISSUE #752: Groovy Filtering issues (and Dynamic configuration)
Test filtering isn't working as expected when using gradle plugin with groovy scripts

I'm trying to configure filteringConfiguration in my groovy scripts using Wrapper class' parameters:

```
  filteringConfiguration {
    allowlist {
      annotationFilter = findProperty('annotationFilter')?.split(',') ?: ['androidx.test.filters.MediumTest']// only default works, set from property it launches all tests
      testMethodFilter = findProperty('testMethodFilter')?.tokenize(',') ?: ['failedRestaurantOrder']//doesn't work in both variants - Scheduling 0 tests
      fullyQualifiedClassnameFilter = ['com.package.AgreementEspressoTest', 'com.package.PromosOnSummaryEspressoTest']//doesn't work - Scheduling 0 tests
      simpleClassNameFilter = ['AgreementEspressoTest', 'PromosOnSummaryEspressoTestKt']//doesn't work - Scheduling 0 tests
    }
  }
```
some of them seem to not work at all (simpleClassNameFilter, fullyQualifiedClassnameFilter, testMethodFilter), some are missing (fully-qualified-test-name), none working if set from property as a list - I've tried split(',') and tokenize(',') to no avail

What I did:
```
task1:
'marathonDebugAndroidTest -PannotationFilter=androidx.test.filters.MediumTest'
config:
    allowlist {
      annotationFilter = findProperty('annotationFilter')?.split(',') ?: ['androidx.test.filters.LargeTest']
      println(annotationFilter)//[androidx.test.filters.MediumTest]
    }
result: 
I 14:07:49.615 [main @coroutine#1] <com.malinskiy.marathon.Marathon> Scheduling 487 tests (all tests - incorrect)
```

```
task2:
'marathonDebugAndroidTest -PtestMethodFilter=failedRestaurantOrder,checkPromoBlockDisabledExp
config:
    allowlist {
      testMethodFilter = findProperty('testMethodFilter')?.tokenize(',') ?: []
      println(testMethodFilter)//[failedRestaurantOrder, checkPromoBlockDisabledExp]

//      testMethodFilter = ['failedRestaurantOrder', 'checkPromoBlockDisabledExp']//also doesn't work
    }
result: 
I 13:51:44.318 [main @coroutine#1] <com.malinskiy.marathon.Marathon> Scheduling 0 tests (incorrect, should be 2)
```

also I've tried fully qualified test names, tried to add Kt to class names, but nothing worked

ildar2:
fix #752
fixed Wrapper.toList() so it would fill 'values' for FilterConfigurations if there are more than 1 entry. otherwise it would remain as 'regex'
added a 'fully-qualified-test-name' implementation


ildar2:
@Malinskiy let me know if there is something to fix in this pr. I've thought about implementing `file` filtering, but decided not to stuff it in there. 
There are two options I've considered for `file`:
1. check if there is a `"file:"` prefix in filtering config
2. create a separate config for `file`

both have their trade offs, but I like the first one better

Malinskiy:
@ildar2 thanks a lot for the PR! I'll have time to review this during the weekend, will reply back here if something needs to be changed. From a quick glance - there should be something added to the docs about the behavior here with mapping single filter parameter to regex and multiple parameters to values. I wouldn't bother with file filters since gradle community seems to insist everything should be in `xxx.gradle`.

ildar2:
Hey! How's it going? We pretty much got around this with "test1|test2" regex hack, but it would be easier for others to use marathon with this addition

Malinskiy (approved):

PR #755, opened 2023-02-16
fix(ios): reuse apple test bundle identifier

Malinskiy:


vendor/vendor-ios/src/main/kotlin/com/malinskiy/marathon/ios/AppleVendor.kt
```diff
@@ -34,6 +34,8 @@ val AppleVendor = module {
     }
     single<TestParser?> { AppleTestParser(get(), get(), get()) }
     single<MarathonLogConfigurator> { AppleLogConfigurator(get()) }
-    single<TestBundleIdentifier?> { AppleTestBundleIdentifier() }
-    single<AppleTestBundleIdentifier> { AppleTestBundleIdentifier() }
```
Malinskiy, line 38: This mistakenly created 2 bundle identifiers and hence produced "no test bundle identifier" errors in runtime

PR #756, opened 2023-02-21
feat(ios): extra artifacts

Malinskiy:


Malinskiy:
Should be reworked as fileSyncConfiguration for Android

PR #758, opened 2023-03-04
feat(core): execution strategy

Malinskiy:
Removes `strictMode` in favor of `executionMode` consisting of:
`mode`:
**ANY_SUCCESS** test passes if any of executions is passing this mode works only if there is no complex sharding strategy applied

Why: it doesn't make sense when user asks for N number of tests to run explicitly, and we pass on the first one. Sharding used to verify probability of passing with an explicit boundary for precision

**ALL_SUCCESS** test passes if and only if all the executions are passing this mode works only if there are no retries, i.e. no complex flakiness strategy, no retry strategy

Why: when adding retries to tests with retry+flakiness strategies users want to trade-off cost for reliability, i.e. add more retries
and pass if one of them passes, so retries only make sense for the [ANY_SUCCESS] mode. When we use [ALL_SUCCESS] mode it means user wants to verify each test with a number of tries (they are not retries per se) and pass only if all of them succeed. This is the case when fixing a flaky test or adding a new test, and we want to have a signal that the test is fixed/not flaky.

`fast`: fail-fast or success-fast depending on the value of [mode]. This doesn't affect the result of the run, it only saves compute time


vendor/vendor-android/src/main/kotlin/com/malinskiy/marathon/android/executor/listeners/TestResultsListener.kt
```diff
@@ -18,18 +18,18 @@ import com.malinskiy.marathon.test.toTestName
...
     private val deferred: CompletableDeferred<TestBatchResults>,
     private val timer: Timer,
-    private val progressReporter: ProgressReporter,
     private val poolId: DevicePoolId,
     attachmentProviders: List<AttachmentProvider>,
     private val attachmentCollector: AttachmentCollector = AttachmentCollector(attachmentProviders),
 ) : AccumulatingResultTestRunListener(timer), AttachmentListener by attachmentCollector {
 
     private val logger = MarathonLogging.logger("TestRunResultsListener")
+    private val progressReporter = ProgressReporter(testBatch, poolId, device.toDeviceInfo())
```
tagantroy: do we need this? I see that you dropped line 102 and looks like it is unused now

core/src/main/kotlin/com/malinskiy/marathon/execution/progress/PoolProgressAccumulator.kt
```diff
@@ -0,0 +1,426 @@
...
+            on<TestEvent.AddRetry> {
+                transitionTo(TestState.Passed(total = total + 1, running = running, done = done))
+            }
+            on<TestEvent.RemoveAttempts> {
+                transitionTo(TestState.Passed(total = total - it.count, running = running, done = done))
+            }
+        }
+        onTransition {
+            if (it as? StateMachine.Transition.Valid !is StateMachine.Transition.Valid) {
+                logger.error { "from ${it.fromState} event ${it.event}" }
```
tagantroy, line 250: Would be nice to add more context here

core/src/main/kotlin/com/malinskiy/marathon/execution/Scheduler.kt
```diff
@@ -32,26 +32,25 @@ import kotlin.coroutines.CoroutineContext
...
     private val job = Job()
-    private val pools = ConcurrentHashMap<DevicePoolId, SendChannel<FromScheduler>>()
+    private val pools = ConcurrentHashMap<DevicePoolId, DevicePoolActor>()
+    private val results = ConcurrentHashMap<DevicePoolId, PoolProgressAccumulator>()
     private val poolingStrategy = configuration.poolingStrategy.toPoolingStrategy()
 
     private val logger = MarathonLogging.logger("Scheduler")
 
-    suspend fun execute() {
+    suspend fun execute() : Boolean {
```
sergioboi, line 56: It's not clear what this `bool` means

core/src/main/kotlin/com/malinskiy/marathon/execution/progress/PoolProgressAccumulator.kt
```diff
@@ -0,0 +1,426 @@
...
+        println("${toPercent(progress())} | [${poolId.name}]-[${device.serialNumber}] ${test.toTestName()} started")
+    }
+
+    /**
+     * @param final used for incomplete tests to signal no more retries left, hence a decision on the status has to be made
+     */
+    fun testEnded(device: DeviceInfo, testResult: TestResult, final: Boolean = false): TestAction? {
+        return when (testResult.status) {
+            TestStatus.FAILURE -> {
+                println("${toPercent(progress())} | [${poolId.name}]-[${device.serialNumber}] ${testResult.test.toTestName()} failed")
```
sergioboi, line 277: Replace with logger?

core/src/main/kotlin/com/malinskiy/marathon/execution/progress/PoolProgressAccumulator.kt
```diff
@@ -0,0 +1,426 @@
...
+                tests[testResult.test.toTestName()]?.transition(TestEvent.Incomplete(device, testResult, final)).sideffect()
+            }
+        }
+    }
+
+    /**
+     * Should always be called before testEnded, otherwise the FSM might transition into a terminal state prematurely
+     */
+    fun retryTest(test: Test): TestAction? {
+        return tests[test.toTestName()]?.transition(TestEvent.AddRetry).sideffect()
```
sergioboi: Should we do warn log when tests returns `null`?

core/src/main/kotlin/com/malinskiy/marathon/execution/progress/ProgressReporter.kt
```diff
@@ -1,82 +1,25 @@
...
-        execute(poolId) { it.addTestDiscoveredDuringRuntime(test) }
+class ProgressReporter(private val batch: TestBatch, private val poolId: DevicePoolId, private val device: DeviceInfo) {
+    fun testStarted(test: Test) {
+        println("${batch.id} | [${poolId.name}]-[${device.serialNumber}] ${test.toTestName()} started")
     }
 
-    fun addRetries(poolId: DevicePoolId, count: Int) {
-        execute(poolId) { it.addTestRetries(count) }
+    fun testFailed(test: Test) {
+        println("${batch.id} | [${poolId.name}]-[${device.serialNumber}] ${test.toTestName()} failed")
```
sergioboi, line 15: Logger?

core/src/main/kotlin/com/malinskiy/marathon/execution/queue/TestAction.kt
```diff
@@ -1,8 +1,5 @@
 package com.malinskiy.marathon.execution.queue
 
-import com.malinskiy.marathon.device.DeviceInfo
-import com.malinskiy.marathon.execution.TestResult
-
 sealed class TestAction {
-    data class SaveReport(val deviceInfo: DeviceInfo, val testResult: TestResult) : TestAction()
+    object Conclude : TestAction()
```
sergioboi: Maybe use most common name for that action. `End` or `Complete`

core/src/main/kotlin/com/malinskiy/marathon/execution/queue/TestState.kt
```diff
@@ -1,24 +1,36 @@
...
+        val total: Int,
+        val running: Int,
+        val done: Int,
+    ) : TestState()
 
-    data class Executed(
-        val device: DeviceInfo,
-        val testResult: TestResult,
-        val count: Int
+    data class Failing(
```
sergioboi, line 19: Right now it's not fully cleared difference between Failed and Failing.
Maybe we could use additional differentiation: `Terminal` States and `NotTerminal`

configuration/src/test/resources/fixture/config/sample_1.yaml
```diff
@@ -68,7 +68,9 @@ excludeSerialRegexes:
   - "emulator-5002"
 ignoreFailures: false
 isCodeCoverageEnabled: false
-strictMode: true
+executionStrategy:
+  mode: ANY_SUCCESS
+  fast: true
```
VictorIreri, line 73: Will you be updating the documentation in this PR?

configuration/src/main/kotlin/com/malinskiy/marathon/config/strategy/ExecutionStrategyConfiguration.kt
```diff
@@ -0,0 +1,31 @@
...
+
+import com.fasterxml.jackson.annotation.JsonProperty
+
+/**
+ * @property fast fail-fast or success-fast depending on the value of [mode]. This doesn't affect the result
+ *  of the run, it only saves compute time
+ */
+data class ExecutionStrategyConfiguration(
+    @JsonProperty("mode") val mode: ExecutionMode = ExecutionMode.ANY_SUCCESS,
+    @JsonProperty("fast") val fast: Boolean = true,
```
VictorIreri, line 11: Given it doesn't affect the result of the run, why is `fast` configurable by the user?

configuration/src/main/kotlin/com/malinskiy/marathon/config/strategy/ExecutionStrategyConfiguration.kt
```diff
@@ -0,0 +1,31 @@
...
+ *  
+ *  Why: it doesn't make sense when user asks for N number of tests
+ *  to run explicitly, and we pass on the first one. Sharding used to verify probability of passing with an explicit boundary for precision
+ * @property ALL_SUCCESS test passes if and only if all the executions are passing
+ *  this mode works only if there are no retries, i.e. no complex flakiness strategy, no retry strategy
+ *  
+ *  Why: when adding retries to tests with retry+flakiness strategies users want to trade-off cost for reliability, i.e. add more retries
+ *  and pass if one of them passes, so retries only make sense for the [ANY_SUCCESS] mode. When we use [ALL_SUCCESS] mode it means user
+ *  wants to verify each test with a number of tries (they are not retries per se) and pass only if all of them succeed. This is the case
+ *  when fixing a flaky test or adding a new test, and we want to have a signal that the test is fixed/not flaky.
```
VictorIreri, line 26: Is there a file where the full configuration defined by the user is parsed and potentially conflicting configurations such as the ones described here are identified? If so, perhaps this explanation belongs there. Furthermore, would such conflicts be presented to the user at runtime, either as a warning or error?

configuration/src/main/kotlin/com/malinskiy/marathon/config/Configuration.kt
```diff
@@ -142,37 +143,37 @@ data class Configuration private constructor(
...
-
-        var screenRecordingPolicy: ScreenRecordingPolicy = ScreenRecordingPolicy.ON_FAILURE,
-
-        var analyticsTracking: Boolean = false,
-        var deviceInitializationTimeoutMillis: Long = DEFAULT_DEVICE_INITIALIZATION_TIMEOUT_MILLIS,
-
-        var outputConfiguration: OutputConfiguration = OutputConfiguration(),
-        var vendorConfiguration: VendorConfiguration = VendorConfiguration.EmptyVendorConfiguration(),
+         val name: String,
+         val outputDir: File,
```
VictorIreri, line 148: Doesn't really matter but should these spaces have been added?
Also, I think there should be a new line above this class between lines 144 & 145.

core/src/main/kotlin/com/malinskiy/marathon/execution/progress/PoolProgressAccumulator.kt
```diff
@@ -0,0 +1,426 @@
...
+        }
+    }
+
+    fun progress(): Float {
+        var done = 0
+        var total = 0
+
+        tests.values.forEach {
+            val state = it.state
+            when (state) {
```
VictorIreri, line 381: Today I learned you could do this when the IDE recommended it 🙂
```suggestion
            when (val state = it.state) {
```

core/src/main/kotlin/com/malinskiy/marathon/execution/queue/TestAction.kt
VictorIreri: Or `Finish`?

vendor/vendor-android/src/main/kotlin/com/malinskiy/marathon/android/executor/listeners/TestResultsListener.kt
Malinskiy: Great question. So currently printing any of these doesn't mean marathon's core will actually see these events. Only after the batch results are calculated. I think these would confuse users so I'll remove them

Malinskiy: iOS implementation should probably also not print these. If users want this level of information it's accessible using debug mode

core/src/main/kotlin/com/malinskiy/marathon/execution/progress/PoolProgressAccumulator.kt
Malinskiy: At this point such transition is impossible: all states respond to every event

core/src/main/kotlin/com/malinskiy/marathon/execution/Scheduler.kt
Malinskiy: Added return quickdoc

core/src/main/kotlin/com/malinskiy/marathon/execution/progress/PoolProgressAccumulator.kt
Malinskiy: This is intended: it's the default way marathon prints progress when users have debug==false. We don't have an abstraction for this

core/src/main/kotlin/com/malinskiy/marathon/execution/progress/PoolProgressAccumulator.kt
Malinskiy: Good call, reworked

core/src/main/kotlin/com/malinskiy/marathon/execution/progress/ProgressReporter.kt
Malinskiy: Same as above. this is the old mechanism to show progress in stdout. I've removed it since most of the info should come from poolprogressaccumulator now

core/src/main/kotlin/com/malinskiy/marathon/execution/queue/TestAction.kt
Malinskiy: One of the hardest problems in CS :) Let's use 'Complete'

core/src/main/kotlin/com/malinskiy/marathon/execution/queue/TestState.kt
Malinskiy: My impression is that using adjectives ending with `ing` is quite common for this purpose in English i.e. https://buildkite.com/docs/pipelines/defining-steps#build-states

configuration/src/test/resources/fixture/config/sample_1.yaml
Malinskiy: Yes, already updated

configuration/src/main/kotlin/com/malinskiy/marathon/config/strategy/ExecutionStrategyConfiguration.kt
Malinskiy: This is explained in the docs, short version is you might want to run other retries anyways in some scenarios because test status is not the only output of a test, test probability of passing is one of the examples

configuration/src/main/kotlin/com/malinskiy/marathon/config/strategy/ExecutionStrategyConfiguration.kt
Malinskiy: Configuration consists of multiple files, there is no single place to define all of it. Defaults should work for most people, if something is not right marathon will produce logical validation error during runtime. Doing so in yaml is not possible unless we have custom linting for yaml.

I've updated `LogicalConfigurationValidator` to fail in runtime under the edge-cases outlined here

configuration/src/main/kotlin/com/malinskiy/marathon/config/Configuration.kt
Malinskiy: Not everything is reformatted according to the embedded code style. It works as an incremental code style formatter - you change something in a file and once you commit it will have approriate format

Malinskiy: Added a newline

core/src/main/kotlin/com/malinskiy/marathon/execution/progress/PoolProgressAccumulator.kt
Malinskiy: You can also ignore IDE and language features for the sake of simplicity and readability ;)
I prefer to define variables separately from the branching logic

VictorIreri: Agreed.

docs/docs/intro/configure.md
```diff
@@ -375,24 +375,121 @@ marathon {
 </TabItem>
 </Tabs>
 
-### Strict mode
-By default, if one of the test retries succeeds then the test is considered successfully executed. If you require success status only when
-all retries were executed successfully you can enable the strict mode. This may be useful to verify that flakiness of tests was fixed for
-example.
+### Execution strategy
+When executing tests with retries there are multiple trade-offs to be made. Two execution strategies are supported: any success or all success.
+By default, any success strategy is used with fast execution i.e. if one of the test retries succeeds then the test is considered successfully
```
VictorIreri, line 380: Worth changing all instances of "any success" & "all success" to `ANY_SUCCESS` & `ALL_SUCCESS` for easier reading?

docs/docs/intro/configure.md
```diff
@@ -375,24 +375,121 @@ marathon {
...
+executed and all non-started retries are removed.
+
+#### Any success
+Test passes if any of executions is passing. This mode works only if there is no complex sharding strategy applied. This is the default.
```
VictorIreri:
```suggestion
Test passes if any of its executions are passing. This mode works only if there is no complex sharding strategy applied. This is the default.
```

docs/docs/intro/configure.md
```diff
@@ -375,24 +375,121 @@ marathon {
...
+</Tabs>
+
+:::info
+
+Complex sharding with any success mode doesn't make sense when user asks for N number of tests to run explicitly, and we pass on the first one.
+
+:::
+
+#### All success
+Test passes if and only if all the executions are passing. This mode works only if there are no retries, i.e. no complex flakiness strategy, no retry strategy.
```
VictorIreri:
```suggestion
Test passes if and only if all its executions are passing. This mode works only if there are no retries, i.e. no complex flakiness strategy, no retry strategy.
```

docs/docs/intro/configure.md
```diff
@@ -375,24 +375,121 @@ marathon {
...
+  executionStrategy = ExecutionStrategyConfiguration(ExecutionMode.ALL_SUCCESS)
+}
+```
+
+</TabItem>
+</Tabs>
+
+:::info
+
+Adding retries with retry/flakiness strategies means users wants to trade-off cost for reliability, i.e. add more retries and pass if one
```
VictorIreri:
```suggestion
Adding retries with retry/flakiness strategies means users want to trade off cost for reliability, i.e. add more retries and pass if one 
```

Also, Is execution time the only "cost" being referred to here and elsewhere on this page? If so, perhaps we can be more specific.

docs/docs/intro/configure.md
```diff
@@ -375,24 +375,121 @@ marathon {
...
+
+</TabItem>
+</Tabs>
+
+:::info
+
+Adding retries with retry/flakiness strategies means users wants to trade-off cost for reliability, i.e. add more retries and pass if one 
+of test retries passes, so retries only make sense for any success mode. 
+
+When we use all success mode it means user wants to verify each test with a number of tries (they are not retries per se) and pass only if
```
VictorIreri:
```suggestion
When we use all success mode it means the user wants to verify each test with a number of tries (they are not retries per se) and pass only if 
```

docs/docs/intro/configure.md
```diff
@@ -375,24 +375,121 @@ marathon {
...
+Adding retries with retry/flakiness strategies means users wants to trade-off cost for reliability, i.e. add more retries and pass if one 
+of test retries passes, so retries only make sense for any success mode. 
+
+When we use all success mode it means user wants to verify each test with a number of tries (they are not retries per se) and pass only if 
+all of them succeed. This is the case when fixing a flaky test or adding a new test, and we want to have a signal that the test is fixed/not flaky.
+
+:::
+
+#### Fast execution mode
+When test reaches a state when a decision about it's state can be made marathon can remove additional in-progress retries.
```
VictorIreri:
```suggestion
When the test reaches a state where a decision about its state can be made, marathon can remove additional in-progress retries.
```

configuration/src/main/kotlin/com/malinskiy/marathon/config/strategy/ExecutionStrategyConfiguration.kt
VictorIreri: Just read it. Totally makes sense to me. 👍

docs/docs/intro/configure.md
```diff
@@ -375,24 +375,121 @@ marathon {
...
+This decision point is different depending on the execution mode used. Let's walk through two examples.
+
+Assume any success strategy is used and 100 retries are scheduled for a test A via flakiness strategy. Let's say first 3 failed and the 4th attempt succeeded. At 
+this point the test should already be considered passed since any success out of all retries leads to the result by definition of any successs
```
VictorIreri, line 468: Except for the first use of "any success" on this line, is it worth changing all instances of "any success" & "all success" to `ANY_SUCCESS` & `ALL_SUCCESS` for easier reading?

docs/docs/intro/configure.md
```diff
@@ -375,24 +375,121 @@ marathon {
...
+When we use all success mode it means user wants to verify each test with a number of tries (they are not retries per se) and pass only if 
+all of them succeed. This is the case when fixing a flaky test or adding a new test, and we want to have a signal that the test is fixed/not flaky.
+
+:::
+
+#### Fast execution mode
+When test reaches a state when a decision about it's state can be made marathon can remove additional in-progress retries.
+This decision point is different depending on the execution mode used. Let's walk through two examples.
+
+Assume any success strategy is used and 100 retries are scheduled for a test A via flakiness strategy. Let's say first 3 failed and the 4th attempt succeeded. At
```
VictorIreri, line 467:
```suggestion
Assume the any success strategy is used and 100 retries are scheduled for test A via the flakiness strategy. Let's say the first 3 failed and the 4th attempt succeeded. At 
```

docs/docs/intro/configure.md
```diff
@@ -375,24 +375,121 @@ marathon {
...
+
+:::
+
+#### Fast execution mode
+When test reaches a state when a decision about it's state can be made marathon can remove additional in-progress retries.
+This decision point is different depending on the execution mode used. Let's walk through two examples.
+
+Assume any success strategy is used and 100 retries are scheduled for a test A via flakiness strategy. Let's say first 3 failed and the 4th attempt succeeded. At 
+this point the test should already be considered passed since any success out of all retries leads to the result by definition of any successs
+execution strategy. To save cost one can remove additional non-started retries by using fast execution mode (this is the default behavior for
```
VictorIreri, line 469: Is "time" the only cost being referred to here and elsewhere on this page? If so, perhaps "time" is more descriptive.

docs/docs/intro/configure.md
```diff
@@ -375,24 +375,121 @@ marathon {
...
+any success strategy). On the other hand one could disable fast execution and get much more accurate statistics about this test by executing
+more retries and calculating the probability of passing as a measure of flakiness for test A.
+
+Assume all success strategy is used and 100 retries are scheduled using sharding strategy. Let's say first 3 passed and the 4th attempt failed.
```
VictorIreri, line 473:
```suggestion
Assume the all success strategy is used and 100 retries are scheduled using the sharding strategy. Let's say the first 3 passed and the 4th attempt failed.
```

configuration/src/main/kotlin/com/malinskiy/marathon/config/strategy/ExecutionStrategyConfiguration.kt
VictorIreri: Just read it. Totally makes sense to me. 👍

docs/docs/intro/configure.md
VictorIreri: Except for the first use of "any success" on line 468, is it worth changing all instances of "any success" & "all success" in this doc to `ANY_SUCCESS` & `ALL_SUCCESS` for easier reading?

docs/docs/intro/configure.md
Malinskiy: Cost is how much one pays with $ for the test runs

docs/docs/intro/configure.md
Malinskiy: replaced

docs/docs/intro/configure.md
Malinskiy: done

docs/docs/intro/configure.md
Malinskiy: changed the wording as suggested

docs/docs/intro/configure.md
Malinskiy: done

docs/docs/intro/configure.md
Malinskiy: done

docs/docs/intro/configure.md
Malinskiy: done

PR #776, opened 2023-03-20
fix(core): SingleValueTestFilter if file not exist then should be evaluated as empty

Vacxe:
…te as empty

### DIFF
--- a/core/src/main/kotlin/com/malinskiy/marathon/execution/filter/SingleValueTestFilter.kt
+++ b/core/src/main/kotlin/com/malinskiy/marathon/execution/filter/SingleValueTestFilter.kt
@@ -1,5 +1,6 @@
 package com.malinskiy.marathon.execution.filter
 
+import com.malinskiy.marathon.execution.TestFilter
 import com.malinskiy.marathon.log.MarathonLogging
 import com.malinskiy.marathon.test.Test
 import java.io.File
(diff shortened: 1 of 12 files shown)


core/src/main/kotlin/com/malinskiy/marathon/execution/filter/SingleValueTestFilter.kt
```diff
@@ -10,12 +10,15 @@ open class SingleValueTestFilter(
...
 ) : TestFilter {
+    private val log = MarathonLogging.logger("SingleValueTestFilter")
+
     private val fileValuesCache: List<String>? by lazy {
         file?.let { valuesFile ->
             if (valuesFile.exists()) {
                 valuesFile.readLines().filter { it.isNotBlank() }
             } else {
-                null
+                log.error { "Filtering configuration file ${valuesFile.absoluteFile} is not exist. Applying empty list." }
```
Malinskiy: is not exist -> does not exist

core/src/main/kotlin/com/malinskiy/marathon/execution/filter/TestFilter.kt
```diff
@@ -1,4 +1,4 @@
-package com.malinskiy.marathon.execution
```
Malinskiy, line 1: This is going to break gradle plugin users. Please don't do this

Malinskiy: This is going to break gradle plugin users. Please don't do this

Malinskiy (changes requested):

Malinskiy (approved):

PR #777, opened 2023-03-21
feat(core): class named batching strategy

Vacxe:
Very controversial thing from "Clean UI tests paradigm". However can be used by somebody.

### DIFF
--- a/core/src/main/kotlin/com/malinskiy/marathon/execution/strategy/impl/batching/ClassNameBatchingStrategy.kt
+++ b/core/src/main/kotlin/com/malinskiy/marathon/execution/strategy/impl/batching/ClassNameBatchingStrategy.kt
@@ -1,0 +1,35 @@
+package com.malinskiy.marathon.execution.strategy.impl.batching
+
+import com.malinskiy.marathon.analytics.external.Analytics
+import com.malinskiy.marathon.config.strategy.BatchingStrategyConfiguration
+import com.malinskiy.marathon.execution.bundle.TestBundleIdentifier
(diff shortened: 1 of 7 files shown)


Malinskiy:
Needs documentation update within the same PR

core/src/main/kotlin/com/malinskiy/marathon/execution/strategy/impl/batching/ClassNameBatchingStrategy.kt
```diff
@@ -0,0 +1,35 @@
...
+import com.malinskiy.marathon.execution.bundle.TestBundleIdentifier
+import com.malinskiy.marathon.execution.strategy.BatchingStrategy
+import com.malinskiy.marathon.test.Test
+import com.malinskiy.marathon.test.TestBatch
+import java.util.Queue
+
+class ClassNameBatchingStrategy(private val cnf: BatchingStrategyConfiguration.ClassNameBatchingStrategyConfiguration) : BatchingStrategy {
+
+    override fun process(queue: Queue<Test>, analytics: Analytics, testBundleIdentifier: TestBundleIdentifier?): TestBatch {
+        val firstGroup = queue.toList().groupBy { test -> "${test.pkg}.${test.clazz}" }.values.first()
```
Malinskiy: Do you really need to group all of the tests by group all the time? For large queue this can be a problem.
I'd suggest getting a single key and then making a single pass over the queue to find tests belonging to that single key.

Malinskiy (changes requested):

Malinskiy (approved):

PR #779, opened 2023-03-22
fix(core): multiple allow lists confings

Vacxe:
Update filtering logic


Vacxe:
Will not works properly with "fragmentation" configuration

PR #809, opened 2023-06-27
Marathon new version

abensamal:


Malinskiy:
Glad you found the [documented option](https://docs.marathonlabs.io/configuration/filtering/#values-file-filtering) to do exactly this @abensamal :)

PR #814, opened 2023-07-07
fix(ios): propagate test status to skip successful tests

Malinskiy:
ON_FAILURE screen recording policy should propagate test status to skip successful tests


Malinskiy:
CircleCI integration was setup automatically due to org settings, can safely ignore it

PR #820, opened 2023-07-25
Add support for bugsnagReporting in gradle config.

tabletenniser:
See https://github.com/MarathonLabs/marathon/issues/818. This PR essentially adds feature parity between `analyticsTracking` and `bugsnagReporting`

This PR also:
* Fix a bug where we still track analytics even if `analyticsTracking` is set to False in the gradle config
* Fix an incorrect CLI helper message on analyticsTracking where the default value is now to track analytics

Things are a bit messy today because `analyticsTracking` can be set in both Marathonfile / gradle config and CLI today. In the case where we run Marathon from gradle, we can only set it inside gradle config but not the CLI. Before this PR, `analyticsTracking` defaults to False for Marathon config but defaults to True for CLI. This makes it impossible to disable `analyticsTracking` when running Marathon from Gradle because the CLI's default True value will always take priority and overwrite whatever value we have in the Marathonfile / gradle config. This PR changes it such that we default to True for both Marathonfile / gradle config and CLI, and if we set it to False explicitly in either, we will disable the analytics tracking. A similar pattern is also applied to `bugsnagReporting`.

This PR is tested by:
* End to end using the `samples/android-app` sample app where we see `Noop exceptions reporter started` with `bugsnagReporting` set to False in the gradle config.
* Unit tests on `ConfigurationFactory.kt`

### DIFF
--- a/cli/src/main/kotlin/com/malinskiy/marathon/cli/ApplicationView.kt
+++ b/cli/src/main/kotlin/com/malinskiy/marathon/cli/ApplicationView.kt
@@ -3,6 +3,7 @@ package com.malinskiy.marathon.cli
 import com.github.ajalt.clikt.core.PrintMessage
 import com.github.ajalt.clikt.core.subcommands
 import com.malinskiy.marathon.Marathon
+import com.malinskiy.marathon.exceptions.ExceptionsReporter
 import com.malinskiy.marathon.android.AndroidVendor
 import com.malinskiy.marathon.android.adam.di.adamModule
 import com.malinskiy.marathon.cli.args.CliConfiguration
@@ -45,10 +46,8 @@ private fun execute(cliConfiguration: CliConfiguration) {
         else -> throw IllegalArgumentException("Please handle the new format of cliConfiguration=$cliConfiguration")
     }
 
-    val bugsnagExceptionsReporter = ExceptionsReporterFactory.get(marathonStartConfiguration.bugsnagReporting)
+    var bugsnagExceptionsReporter: ExceptionsReporter? = null;
     try {
-        bugsnagExceptionsReporter.start(AppType.CLI)
-
         logger.info { "Checking ${marathonStartConfiguration.marathonfile} config" }
         if (!marathonStartConfiguration.marathonfile.isFile) {
             logger.error { "No config ${marathonStartConfiguration.marathonfile.absolutePath} present" }
@@ -58,7 +57,12 @@ private fun execute(cliConfiguration: CliConfiguration) {
         val configuration = ConfigurationFactory(
             marathonfileDir = marathonStartConfiguration.marathonfile.canonicalFile.parentFile,
             analyticsTracking = marathonStartConfiguration.analyticsTracking,
+            bugsnagReporting = marathonStartConfiguration.bugsnagReporting,
         ).parse(marathonStartConfiguration.marathonfile)
+
+        bugsnagExceptionsReporter = ExceptionsReporterFactory.get(configuration.bugsnagReporting)
+        bugsnagExceptionsReporter.start(AppType.CLI)
+
         val vendorConfiguration = configuration.vendorConfiguration
         val modules = when (vendorConfiguration) {
             is VendorConfiguration.IOSConfiguration -> {
@@ -87,6 +91,8 @@ private fun execute(cliConfiguration: CliConfiguration) {
         }
     } finally {
         stopKoin()
-        bugsnagExceptionsReporter.end()
+        if (bugsnagExceptionsReporter != null) {
+            bugsnagExceptionsReporter.end()
+        }
     }
 }
--- a/configuration/src/main/kotlin/com/malinskiy/marathon/config/Configuration.kt
+++ b/configuration/src/main/kotlin/com/malinskiy/marathon/config/Configuration.kt
@@ -173,8 +173,8 @@ data class Configuration private constructor(
 
          var screenRecordingPolicy: ScreenRecordingPolicy = ScreenRecordingPolicy.ON_FAILURE,
 
-         var analyticsTracking: Boolean = false,
-         var bugsnagReporting: Boolean = false,
+         var analyticsTracking: Boolean = true,
+         var bugsnagReporting: Boolean = true,
          var deviceInitializationTimeoutMillis: Long = DEFAULT_DEVICE_INITIALIZATION_TIMEOUT_MILLIS,
 
          var outputConfiguration: OutputConfiguration = OutputConfiguration(),
--- a/configuration/src/main/kotlin/com/malinskiy/marathon/config/Configuration.kt
+++ b/configuration/src/main/kotlin/com/malinskiy/marathon/config/Configuration.kt
@@ -49,6 +49,7 @@ data class Configuration private constructor(
     val vendorConfiguration: VendorConfiguration,
 
     val analyticsTracking: Boolean,
+    val bugsnagReporting: Boolean,
     val deviceInitializationTimeoutMillis: Long,
 ) {
     fun toMap() =
@@ -109,6 +110,7 @@ data class Configuration private constructor(
         if (screenRecordingPolicy != other.screenRecordingPolicy) return false
         if (vendorConfiguration != other.vendorConfiguration) return false
         if (analyticsTracking != other.analyticsTracking) return false
+        if (bugsnagReporting != other.bugsnagReporting) return false
         if (deviceInitializationTimeoutMillis != other.deviceInitializationTimeoutMillis) return false
 
         return true
@@ -139,6 +141,7 @@ data class Configuration private constructor(
         result = 31 * result + screenRecordingPolicy.hashCode()
         result = 31 * result + vendorConfiguration.hashCode()
         result = 31 * result + analyticsTracking.hashCode()
+        result = 31 * result + bugsnagReporting.hashCode()
         result = 31 * result + deviceInitializationTimeoutMillis.hashCode()
         return result
     }
@@ -171,6 +174,7 @@ data class Configuration private constructor(
          var screenRecordingPolicy: ScreenRecordingPolicy = ScreenRecordingPolicy.ON_FAILURE,
 
          var analyticsTracking: Boolean = false,
+         var bugsnagReporting: Boolean = false,
          var deviceInitializationTimeoutMillis: Long = DEFAULT_DEVICE_INITIALIZATION_TIMEOUT_MILLIS,
 
          var outputConfiguration: OutputConfiguration = OutputConfiguration(),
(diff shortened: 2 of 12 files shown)


cli/src/main/kotlin/com/malinskiy/marathon/cli/ApplicationView.kt
```diff
@@ -45,10 +46,8 @@ private fun execute(cliConfiguration: CliConfiguration) {
         else -> throw IllegalArgumentException("Please handle the new format of cliConfiguration=$cliConfiguration")
     }
 
-    val bugsnagExceptionsReporter = ExceptionsReporterFactory.get(marathonStartConfiguration.bugsnagReporting)
```
Malinskiy, line 48: Any crash reporting should start at the beginning of the code to check as many problems as possible. One example was the classpath incompatibility between gradle's jackson and marathon's jackson which failed to even parse config. This issue is not applicable anymore, but the reasoning for it still applies.

configuration/src/main/kotlin/com/malinskiy/marathon/config/Configuration.kt
```diff
@@ -109,6 +110,7 @@ data class Configuration private constructor(
         if (screenRecordingPolicy != other.screenRecordingPolicy) return false
         if (vendorConfiguration != other.vendorConfiguration) return false
         if (analyticsTracking != other.analyticsTracking) return false
+        if (bugsnagReporting != other.bugsnagReporting) return false
```
Malinskiy, line 113: Good catch

core/src/main/kotlin/com/malinskiy/marathon/Marathon.kt
```diff
@@ -181,8 +181,6 @@ class Marathon(
     }
 
     private fun logSystemInformation() {
-        log.info { "System Information:" }
```
Malinskiy: How is this relevant to the PR?

marathon-gradle-plugin/src/main/kotlin/com/malinskiy/marathon/gradle/MarathonExtension.kt
```diff
@@ -221,6 +221,11 @@ open class MarathonExtension {
      */
     var analyticsTracking: Boolean = true
 
+    /**
+     * Whether to report crashes to Bugsnag. By default, this is enabled. Use **false** to disable.
+     */
+    var bugsnagReporting: Boolean = true
```
Malinskiy, line 227: :+1:

sample/android-app/app/build.gradle.kts
```diff
@@ -51,6 +51,8 @@ android {
 }
 
 marathon {
+    analyticsTracking = false
```
Malinskiy: Please remove these from the sample app. Instead, maybe the docs need an example of disabling bugsnag for both CLI and for Gradle, currently there are none.

configuration/src/main/kotlin/com/malinskiy/marathon/config/serialization/ConfigurationFactory.kt
```diff
@@ -106,10 +107,13 @@ class ConfigurationFactory(
...
                 is VendorConfiguration.EmptyVendorConfiguration -> throw ConfigurationException("No vendor configuration specified")
             }
-
             return configuration.copy(
                 vendorConfiguration = vendorConfiguration,
-                analyticsTracking = analyticsTracking ?: configuration.analyticsTracking
+                // Default value for analyticsTracking / bugsnagReporting in both CLI and marathon / gradle config is true.
+                // If analyticsTracking / bugsnagReporting is set to false in either CLI or marathon / gradle config, it will be disabled.
+                // Note that it's not possible to set the CLI value when running marathon via gradle.
+                analyticsTracking = !(analyticsTracking == false || configuration.analyticsTracking == false),
```
Malinskiy: might be easier to rewrite with De Morgan's law into 
`analyticsTracking == true && configuration.analyticsTracking == true`, same below

Malinskiy (changes requested):

cli/src/main/kotlin/com/malinskiy/marathon/cli/ApplicationView.kt
```diff
@@ -87,6 +91,8 @@ private fun execute(cliConfiguration: CliConfiguration) {
         }
     } finally {
         stopKoin()
-        bugsnagExceptionsReporter.end()
+        if (bugsnagExceptionsReporter != null) {
```
Malinskiy, line 94: Generally if there is a usage of Noop implementation null checks are redundant

cli/src/main/kotlin/com/malinskiy/marathon/cli/ApplicationView.kt
tabletenniser: I think it's a chicken-and-egg problem here. This PR adds the ability to disable bugSnag reporting from the gradle / Marathonfile. But without parsing the Marathon file, we don't know if the setting is disabled.

I guess the best we can do is to always start bugSnag reporting, and if it's disabled in the Marathon file, we terminate it right away. What do you think?

cli/src/main/kotlin/com/malinskiy/marathon/cli/ApplicationView.kt
tabletenniser: This `if condition` is added in case we fail to parse the Marathon file and throw an exception. In this case, `bugsnagExceptionsReporter` isn't initialized yet and we don't want to call `bugsnagExceptionsReporter.end()`

configuration/src/main/kotlin/com/malinskiy/marathon/config/serialization/ConfigurationFactory.kt
tabletenniser: Actually, I don't think it's quite the same because `null` value is allowed for `analyticsTracking`. If `analyticsTracking` is null and `configuration.analyticsTracking` is true, we want it to be enabled (which is what the code currently does). We want to disable it only if one *explicitly* sets it to false for `analyticsTracking` or `configuration.analyticsTracking`.

Though based on my reading of the code, `null` value seems to only be used in unit tests?  Maybe we can make it a mandatory param and dis-allow `null` values?

core/src/main/kotlin/com/malinskiy/marathon/Marathon.kt
tabletenniser: Oops, that was not intentional, added it back.

sample/android-app/app/build.gradle.kts
tabletenniser: Removed and updated `docs/docs/intro/configure.md`

docs/docs/intro/configure.md
```diff
@@ -343,6 +344,42 @@ marathon {
 </TabItem>
 </Tabs>
 
+NOTE: analyticsTracking can also be enabled (default value) / disabled directly from the CLI. It is disabled if it's set to be disabled in either the config or the CLI.
```
Malinskiy: Can you please just use https://docusaurus.io/docs/markdown-features/admonitions?
There are examples in the codebase for them already

docs/docs/intro/configure.md
```diff
@@ -343,6 +344,42 @@ marathon {
...
+```groovy
+marathon {
+  bugsnagReporting = false
+}
+```
+
+</TabItem>
+</Tabs>
+
+NOTE: bugsnagReporting can also be enabled (default value) / disabled directly from the CLI. It is disabled if it's set to be disabled in either the config or the CLI.
```
Malinskiy: Same with admonitions

Malinskiy (changes requested):

Malinskiy (dismissed):

tabletenniser:
Thanks @Malinskiy for the review! Are you going to merge this PR (since I obviously don't have permission to) or do we still need reviews from @tagantroy and @Vacxe ?

Malinskiy:
I'm waiting for one more lgtm for a day and will merge afterwards assuming no issues.

cli/src/main/kotlin/com/malinskiy/marathon/cli/ApplicationView.kt
```diff
@@ -45,10 +46,8 @@ private fun execute(cliConfiguration: CliConfiguration) {
         else -> throw IllegalArgumentException("Please handle the new format of cliConfiguration=$cliConfiguration")
     }
 
-    val bugsnagExceptionsReporter = ExceptionsReporterFactory.get(marathonStartConfiguration.bugsnagReporting)
+    var bugsnagExceptionsReporter: ExceptionsReporter? = null;
```
matzuk, line 49: `lateinit var`?

configuration/src/main/kotlin/com/malinskiy/marathon/config/serialization/ConfigurationFactory.kt
matzuk: To be honest, such constructions are a bit confusing. For me, it is better to write a bit more but make the statement clear. For example,
```
analyticsTracking = 
    if (analyticsTracking == false || configuration.analyticsTracking == false) false 
    then true
```
But this comment may be subjective.

matzuk (dismissed):

configuration/src/main/kotlin/com/malinskiy/marathon/config/serialization/ConfigurationFactory.kt
matzuk: @tabletenniser thank you for your contribution! Approved.
I've added a small comment. Please have a look.

configuration/src/main/kotlin/com/malinskiy/marathon/config/serialization/ConfigurationFactory.kt
```diff
@@ -112,8 +112,8 @@ class ConfigurationFactory(
                 // Default value for analyticsTracking / bugsnagReporting in both CLI and marathon / gradle config is true.
                 // If analyticsTracking / bugsnagReporting is set to false in either CLI or marathon / gradle config, it will be disabled.
                 // Note that it's not possible to set the CLI value when running marathon via gradle.
-                analyticsTracking = !(analyticsTracking == false || configuration.analyticsTracking == false),
-                bugsnagReporting = !(bugsnagReporting == false || configuration.bugsnagReporting == false)
+                analyticsTracking = run {if (analyticsTracking == false || configuration.analyticsTracking == false) false else true},
```
matzuk: why `run` is used here? =)

tabletenniser: Because without it, I'm getting a compilation error: `Type mismatch: inferred type is () -> Boolean but Boolean was expected`  (I'm a completely newbie to Kotlin, so if there's a more proper way to fix the issue, feel free to suggest it!)

matzuk: No worries, I've fixed it for you.

matzuk (approved):

matzuk:
@Malinskiy let's merge

Issue #825, opened 2023-08-02
Erroneous $ in flakiness overhead string


smugleafdev:
**Describe the bug**
There's a random $ hanging out at the end of this string:
Flakiness overhead: 0H 01m 00s$

**To Reproduce**
Steps to reproduce the behaviour:
Just run 0.8.3 and look at the logs at the end for flakiness overhead

**Expected behavior**
Should look like this instead:
Flakiness overhead: 0H 01m 00s

**Additional context**
Here is my comically small commit on my fork, though I won't be offended if you don't use it:
https://github.com/Progressive-Insurance/marathon/commit/1934a9a677e323c98cc4b79cbab51743d3841676

Malinskiy:
Raised in #826

Malinskiy:
In develop, will be release once the new version if ready

PR #826, opened 2023-08-03
fix(core): remove erroneous $ in flakiness overhead

Malinskiy:
Flakiness overhead: 0H 01m 00s$ <- should not be there


Malinskiy:
Closes #825

PR #848, opened 2023-10-17
feat(ios): add xctest parser

Malinskiy:
The iOS test platform has always been problematic to parallelise since there is no way to get a list of tests for execution. Although Swift supports parsing via `--list-tests`, this doesn't work for native iOS targets and not all the tests are written using Swift (some test frameworks merele pretend they're xctest-based)

This change adds xctest parser with support for any xctest-compliant tests including runtime-generated ones. This potentially removes the use-case for testClassRegex since all tests will be reported directly and there is no filtering needed like the filtering for potentially large symbol table of the binary using `nm`.

Note: xctest parser is optional and contains code under EULA


Vacxe:
Great stuff mate

Vacxe (approved):

PR #875, opened 2023-12-08
fix(core): propagate meta properties into final TestResult

matzuk:


core/src/main/kotlin/com/malinskiy/marathon/execution/progress/PoolProgressAccumulator.kt
```diff
@@ -19,11 +19,11 @@ import kotlin.math.roundToInt
 
 class PoolProgressAccumulator(
     private val poolId: DevicePoolId,
-    shard: TestShard,
-    private val configuration: Configuration,
+    val shard: TestShard,
+    val configuration: Configuration,
```
Malinskiy: why the change to public?

Malinskiy (changes requested):

core/src/main/kotlin/com/malinskiy/marathon/execution/progress/PoolProgressAccumulator.kt
matzuk: fixed

Malinskiy: Where? I still see `val configuration: Configuration,`

matzuk: Please check the file. You see the outdated information (GitHub specific).

PR #876, opened 2023-12-11
fix(android_parsing): remote parse without overridden testrun listene…

matzuk:
…r if a user forgot to add the correspondent dependency in the project

### DIFF
--- a/vendor/vendor-android/src/main/kotlin/com/malinskiy/marathon/android/adam/AmInstrumentTestParser.kt
+++ b/vendor/vendor-android/src/main/kotlin/com/malinskiy/marathon/android/adam/AmInstrumentTestParser.kt
@@ -125,7 +125,9 @@ class AmInstrumentTestParser(
                                 testBundleIdentifier.put(test, androidTestBundle)
                             }
 
-                            is TestRunFailed -> Unit
+                            is TestRunFailed -> {
+                                if (overrides.containsKey(LISTENER_ARGUMENT)) throw PossibleListenerIssueException()
+                            }
                             is TestRunStopped -> Unit
                             is TestRunEnded -> Unit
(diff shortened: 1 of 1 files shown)


vendor/vendor-android/src/main/kotlin/com/malinskiy/marathon/android/adam/AmInstrumentTestParser.kt
```diff
@@ -68,7 +78,10 @@ class AmInstrumentTestParser(
 
             val testParserConfiguration = vendorConfiguration.testParserConfiguration
             val overrides: Map<String, String> = when {
-                testParserConfiguration is TestParserConfiguration.RemoteTestParserConfiguration -> testParserConfiguration.instrumentationArgs
+                testParserConfiguration is TestParserConfiguration.RemoteTestParserConfiguration -> {
+                    if (blockListenerArgumentOverride) testParserConfiguration.instrumentationArgs.filterKeys { it != LISTENER_ARGUMENT }
```
Malinskiy: this is going to filter all listener arguments, not just the one for parsing. It's possible to use lots of listeners at the same and just one of them would be the test parser annotation producer

Malinskiy (changes requested):

vendor/vendor-android/src/main/kotlin/com/malinskiy/marathon/android/adam/AmInstrumentTestParser.kt
matzuk: added a check for the value (annotation producer)

matzuk:
closed prior to https://github.com/MarathonLabs/marathon/pull/877

PR #889, opened 2024-02-03
fix(core): add maxFilename limit for outputConfiguration

Malinskiy:
Before:
maxPath limits the overall output file path length
After:
Fine-grained control over max filename and max file path

Reason for change:
Most modern filesystems (ext3, ext4, NTFS, APFS, etc) have max file name of 255 and very large or missing limits for max file path.


configuration/src/main/kotlin/com/malinskiy/marathon/config/OutputConfiguration.kt
```diff
@@ -2,6 +2,10 @@ package com.malinskiy.marathon.config
 
 import com.fasterxml.jackson.annotation.JsonProperty
 
+const val OUTPUT_MAX_PATH = 0
```
matzuk, line 6: It would be better to add the comment: "Value 0 is equivalent to unlimited file path length"

docs/runner/intro/configure.md
```diff
@@ -617,18 +617,24 @@ marathon {
...
+#### Max file name length
 
-By default, the max file path for any output file is capped at 255 characters due to limitations of some OSs. This is the reason why some
-test runs have lots of "File path length cannot exceed" messages in the log. Since there is currently no API to programmatically
-establish this limit, it's the user's responsibility to set a larger value if the OS supports this and the user desires it.
+By default, the max file name length for any output file is capped at 255 characters due to limitations of most file systems. This is the 
+reason why some test runs have lots of "File name length cannot exceed" messages in the log. Since there is currently no API to 
+programmatically establish this limit, it's the user's responsibility to set a larger value if the OS supports this and the user desires it.
+
+:::info
```
matzuk, line 626: I guess it is related to a path not a file name

matzuk (dismissed):

docs/runner/intro/configure.md
Malinskiy: Not really, the new error messages specifically separate filename length and file path length

configuration/src/main/kotlin/com/malinskiy/marathon/config/OutputConfiguration.kt
Malinskiy: Fixed

Issue #55, opened 2018-07-10
Record video/Capture screenshots


tagantroy:

Malinskiy:
+ color(ful) logcat

tagantroy:
#60

PR #895, opened 2024-02-08
feat(ios): proper implementation of output idle timeout

Malinskiy:


vendor/vendor-ios/src/main/kotlin/com/malinskiy/marathon/ios/extensions/Reader.kt
```diff
@@ -77,11 +84,6 @@ fun CoroutineScope.produceLinesManually(
                     } else if (count == -1) {
                         break
```
matzuk, line 85: why don't we close the channel here and in other `break`s?

vendor/vendor-ios/src/main/kotlin/com/malinskiy/marathon/ios/extensions/Reader.kt
```diff
@@ -66,6 +66,13 @@ fun CoroutineScope.produceLinesManually(
                         available > 0 -> inputStream.read(byteArray, 0, min(available, byteArray.size))
                         else -> 0
                     }
+
```
matzuk, line 69: Do we need to catch other exceptions and close the channel on line #55?

Malinskiy: Because there might still be something in the buffer that needs to be flushed to the actual channel. This happens just after the `while` loop

vendor/vendor-ios/src/main/kotlin/com/malinskiy/marathon/ios/extensions/Reader.kt
Malinskiy: Because there might still be something in the buffer that needs to be flushed to the actual channel. This happens just after the while loop

vendor/vendor-ios/src/main/kotlin/com/malinskiy/marathon/ios/extensions/Reader.kt
Malinskiy: According to the documentation `read` can only throw IOException

Malinskiy: inputStream will be closed after flushing because there is `use` around this block

Malinskiy:
Closed in favor of changes in #901

PR #908, opened 2024-03-13
feat(core): add simple-test-name filter

Malinskiy:


docs/runner/configuration/filtering.md
```diff
@@ -107,6 +107,9 @@ Filters tests by their FQTN which is `$package.$class#$method`. The `#` sign is
 ### "fully-qualified-class-name"
 Filters tests by their FQCN which is `$package.$class`                                     |
 
+### "simple-test-name"
```
matzuk, line 110: The name is a bit confusing. I expect to see only the test name. Maybe it is better to call `simple-test-class-and-test-name`?

core/src/main/kotlin/com/malinskiy/marathon/execution/filter/SimpleTestnameFilter.kt
```diff
@@ -0,0 +1,10 @@
+package com.malinskiy.marathon.execution.filter
+
+import com.malinskiy.marathon.config.TestFilterConfiguration
+import com.malinskiy.marathon.test.toSimpleSafeTestName
+
+class SimpleTestnameFilter(cnf: TestFilterConfiguration.SimpleTestnameFilterConfiguration) :
```
matzuk, line 6: I don't see unit tests covering the new filter.

docs/runner/configuration/filtering.md
Malinskiy: Currently marathon supports:
"fully-qualified-test-name" i.e. `package.class#method`
"fully-qualified-class-name" i.e. `package.class`
"simple-class-name" = "fully-qualified-class-name" without the package i.e. `class`

I don't see a problem with the name here since this is a simple form of "fully-qualified-test-name" without the package

matzuk: Your logic is good if you know all the options and you know the reasons behind each name.
My message is that the name should be clear for a developer without the context of other names.
Anyway, it is up to you.

matzuk (approved):

PR #914, opened 2024-03-29
feature(xctestrunEnv): EnvironmentVariables and TestingEnvironment va…

matzuk:
…riables are supported in generated xctestrun

### DIFF
--- a/configuration/src/main/kotlin/com/malinskiy/marathon/config/vendor/VendorConfiguration.kt
+++ b/configuration/src/main/kotlin/com/malinskiy/marathon/config/vendor/VendorConfiguration.kt
@@ -21,6 +21,7 @@ import com.malinskiy.marathon.config.vendor.apple.ios.SigningConfiguration
 import com.malinskiy.marathon.config.vendor.apple.SshConfiguration
 import com.malinskiy.marathon.config.vendor.apple.TimeoutConfiguration as AppleTimeoutConfiguration
 import com.malinskiy.marathon.config.vendor.apple.ios.XcresultConfiguration
+import com.malinskiy.marathon.config.vendor.apple.ios.XctestrunEnvConfiguration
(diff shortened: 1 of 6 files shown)


configuration/src/main/kotlin/com/malinskiy/marathon/config/vendor/VendorConfiguration.kt
```diff
@@ -188,7 +189,7 @@ sealed class VendorConfiguration {
         @JsonProperty("ssh") val ssh: SshConfiguration = SshConfiguration(),
 
         @JsonProperty("xcresult") val xcresult: XcresultConfiguration = XcresultConfiguration(),
-        @JsonProperty("xctestrunEnv") val xctestrunEnv: Map<String, String> = emptyMap(),
+        @JsonProperty("xctestrunEnv") val xctestrunEnv: XctestrunEnvConfiguration = XctestrunEnvConfiguration(),
```
Malinskiy, line 192: Doc should likely change for macos as well

vendor/vendor-apple/base/src/main/kotlin/com/malinskiy/marathon/apple/xctestrun/TestRootFactory.kt
```diff
@@ -209,6 +211,7 @@ class TestRootFactory(
                     arrayOf(
                         TestTarget.withArtifactReinstall(
                             name = bundle.testBundleId,
+                            environmentVariables = xctestrunEnv.appEnvs,
```
Malinskiy: Not sure if empty works here, maybe we should omit this in case of empty envvars

Malinskiy (changes requested):

Malinskiy (approved):

PR #935, opened 2024-05-14
fix(apple): use keepAliveInterval from ssh config

vfadc:
fix the ability to customize keepAliveInterval for ssh connection


Malinskiy (approved):

Malinskiy:
Thanks for the fix @vfadc

PR #937, opened 2024-05-22
feat(core): add filter disabling thought config

Vacxe:


Vacxe:
POC: Don't like `ToggleTestFilter` naming

core/src/main/kotlin/com/malinskiy/marathon/execution/filter/AllureTestFilter.kt
```diff
@@ -20,7 +28,7 @@ class AllureTestFilter(val cnf: TestFilterConfiguration.AllureFilterConfiguratio
         }
     }
 
-    override fun filter(tests: List<Test>): List<Test> {
+    override fun filterPredicate(tests: List<Test>): List<Test> {
```
Malinskiy: Predicate returns `bool`, this function is not a predicate

core/src/main/kotlin/com/malinskiy/marathon/execution/filter/ToggleTestFilter.kt
```diff
@@ -0,0 +1,13 @@
+package com.malinskiy.marathon.execution.filter
+
+import com.malinskiy.marathon.execution.TestFilter
+import com.malinskiy.marathon.test.Test
+
+interface ToggleTestFilter : TestFilter {
```
Malinskiy: Just change the TestFilter to a base class and add this logic there

Malinskiy (approved):

PR #942, opened 2024-06-11
feat(apple): implement importMedia for simulators

vfadc:
Added the ability to push media files to the apple simulator before test run using `xcrun simctl addmedia`
A vendor configuration looks like:
```yaml
importMedia:
  - "media/photo.jpg"
  - "media/video.mp4"
  - "media/contact.vcf"
```

I'm open to rework it in any other configuration ways.

UPDATE: changed the snippet to reflect the final configuration format

### DIFF
--- a/configuration/src/main/kotlin/com/malinskiy/marathon/config/serialization/ConfigurationFactory.kt
+++ b/configuration/src/main/kotlin/com/malinskiy/marathon/config/serialization/ConfigurationFactory.kt
@@ -84,6 +84,9 @@ class ConfigurationFactory(
                             it.testType
                         ).apply { validate() }
                     }
+
+                    val resolvedMediaFiles = iosConfiguration.mediaFiles?.map { ddd -> marathonfileDir.resolve(ddd) }
+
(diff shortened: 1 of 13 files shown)


Malinskiy:
Looking at the help:
```
Add photos, live photos, videos, or contacts to the library of a device.
Usage: simctl addmedia <device> <path> [... <path>]

You can specify multiple files including a mix of photos, videos, and contacts.

You can also specify multiple live photos by providing the photo and video files. They will automatically be discovered and imported correctly.

Contacts support the vCard format.
```

I'm thinking maybe it's better to name the configuration block as `importMedia`, i.e. 
```yaml
importMedia:
  - "media/photo.jpg"
  - "media/video.mp4"
  - "media/contact.vcf
```
WDYT?

Another thing that probably is worth emphasising in the docs is that this import will only be done once per device provisioning, so in the simplest case once during a test run. So if the test manipulates these in any way - it's the test's responsibility to reset it to the state needed.

vfadc:
yes, agree, `importMedia` looks more concise.
Done.
Also added to the doc about side effects of media imports.

Malinskiy (approved):

PR #945, opened 2024-06-14
feat(apple): pulling xcresult policy

vfadc:
Add pulling xcresult policy.
not sure which one would be better, `ON_FAILURE` or `ON_BATCH_FAILURE`, anyway, open for rework

### DIFF
--- a/configuration/src/main/kotlin/com/malinskiy/marathon/config/vendor/apple/ios/XcresultConfiguration.kt
+++ b/configuration/src/main/kotlin/com/malinskiy/marathon/config/vendor/apple/ios/XcresultConfiguration.kt
@@ -3,6 +3,7 @@ package com.malinskiy.marathon.config.vendor.apple.ios
 import com.fasterxml.jackson.annotation.JsonProperty
 
 data class XcresultConfiguration(
+    @JsonProperty("pull") val pull: Boolean? = null,
     @JsonProperty("pullingPolicy") val pullingPolicy: PullingPolicy = PullingPolicy.ALWAYS,
     @JsonProperty("remoteClean") val remoteClean: Boolean = true,
     @JsonProperty("attachments") val attachments: AttachmentsConfiguration = AttachmentsConfiguration(),
--- a/configuration/src/main/kotlin/com/malinskiy/marathon/config/vendor/apple/ios/XcresultConfiguration.kt
+++ b/configuration/src/main/kotlin/com/malinskiy/marathon/config/vendor/apple/ios/XcresultConfiguration.kt
@@ -3,7 +3,7 @@ package com.malinskiy.marathon.config.vendor.apple.ios
 import com.fasterxml.jackson.annotation.JsonProperty
 
 data class XcresultConfiguration(
-    @JsonProperty("pull") val pull: Boolean = true,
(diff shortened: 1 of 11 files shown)


configuration/src/main/kotlin/com/malinskiy/marathon/config/vendor/apple/ios/XcresultConfiguration.kt
```diff
@@ -3,7 +3,7 @@ package com.malinskiy.marathon.config.vendor.apple.ios
 import com.fasterxml.jackson.annotation.JsonProperty
 
 data class XcresultConfiguration(
-    @JsonProperty("pull") val pull: Boolean = true,
+    @JsonProperty("pullingPolicy") val pullingPolicy: PullingPolicy = PullingPolicy.ALWAYS,
```
Malinskiy, line 7: It might be a good idea to save the old option for at least one release and if pull is enabled default to `ALWAYS`. Later we can remove the `pull` option

Malinskiy (changes requested):

configuration/src/main/kotlin/com/malinskiy/marathon/config/vendor/apple/ios/XcresultConfiguration.kt
vfadc: done.
Also I wanted to leave warning deprecation message, but there is no logger in the configuration module. It might be a good idea in future to take it out from the core module.

Malinskiy: There is concurrent work around fancy progress bars in the output, that conflicts with the slf4j logger. Maybe as a next step here we can create our own logger interface and depend on that, but i suspect it will break in gradle plugin

Malinskiy (approved):

matzuk:
Nice job guys!
I just want to emphasize that there is another setting related to artifacts left after an iOS run: [Attachment lifetime](https://docs.marathonlabs.io/runner/apple/configure/ios#attachment-lifetime).
The example is:
```
  xcresult:
    pull: true
    remoteClean: true
    attachments:
      systemAttachmentLifetime: DELETE_ON_SUCCESS
      userAttachmentLifetime: DELETE_ON_SUCCESS
```
Marathon "passes" these params to `xcodebuild` to define desired attachments left after a run.
I guess it makes sense to describe these params and their impact in more details in the documentation.

vfadc:
> Nice job guys! I just want to emphasize that there is another setting related to artifacts left after an iOS run: [Attachment lifetime](https://docs.marathonlabs.io/runner/apple/configure/ios#attachment-lifetime). The example is:
> 
> ```
>   xcresult:
>     pull: true
>     remoteClean: true
>     attachments:
>       systemAttachmentLifetime: DELETE_ON_SUCCESS
>       userAttachmentLifetime: DELETE_ON_SUCCESS
> ```
> 
> Marathon "passes" these params to `xcodebuild` to define desired attachments left after a run. I guess it makes sense to describe these params and their impact in more details in the documentation.

Sure, but these changes don't impact on attachments lifetime, attachment-lifetime params are about storing attachments in xcresult files, but pullingPolicy is about pulling xcresult files. So, I can leave a tip in the doc about it.

PR #950, opened 2024-06-25
fix(core): improve return batch behavior

Malinskiy:


core/src/main/kotlin/com/malinskiy/marathon/execution/device/DeviceActor.kt
```diff
@@ -216,7 +218,9 @@ class DeviceActor(
 
     private fun returnBatchAnd(batch: TestBatch, reason: String, completionHandler: CompletionHandler = {}): Job {
         return launch {
-            pool.send(DevicePoolMessage.FromDevice.ReturnTestBatch(device, batch, reason))
+            withContext(NonCancellable) {
```
Malinskiy, line 221: This makes sure even if the actor job is cancelled the batch will be returned

Malinskiy: By creating a dangling job :disappointed:

PR #977, opened 2024-10-10
fix(ios)/supportsTranscoding-requirement-weakened

danmir:
In our setup, the new Transcoding feature is disabled. By default, `TranscodingConfiguration.enabled == false`. However, in the `AppleSimulatorDevice`, despite this setting, we attempt to run the command `vendorConfiguration.screenRecordConfiguration.videoConfiguration.transcoding.binary --version`. On our CI environment, we do not have `ffmpeg` installed, resulting in a crash with the error message `Caused by: java.io.IOException: Cannot run program "/opt/homebrew/bin/ffmpeg": error=2, No such file or directory`.

To address this issue, I have added an additional check to verify if the feature is enabled.


Malinskiy:
LGTM
thanks @danmir

PR #985, opened 2024-11-21
fix(ios): add xcode timeout result parsing

msdeleyto:
The `TestRunProgressParser` is reading the xcrun logs to detect the end time and calculate the start time. However, when a timeout is defined from XCode and this timeout is triggered, the log looks like:
```
Test Case '-[TestTarget.TestClass test]' exceeded execution time allowance of 3 minutes. The test may have hung; check Xcode's test report for additional diagnostics. If the test requires additional time, use -[XCTestCase executionTimeAllowance] or one of xcodebuild's command-line overrides.
```
Resulting on the parser not being able to correctly identified what happened.

This PR implements an extra case in the `TestRunProgressParser` to match this case and fix the issue reported [here](https://github.com/MarathonLabs/marathon/issues/984)

### DIFF
--- a/vendor/vendor-apple/base/src/main/kotlin/com/malinskiy/marathon/apple/logparser/parser/DeviceFailureParser.kt
+++ b/vendor/vendor-apple/base/src/main/kotlin/com/malinskiy/marathon/apple/logparser/parser/DeviceFailureParser.kt
@@ -4,7 +4,7 @@ import com.malinskiy.marathon.apple.logparser.TestEventProducer
 import com.malinskiy.marathon.apple.test.TestEvent
 import com.malinskiy.marathon.apple.test.TestRunFailed
 
-class DeviceFailureParser : com.malinskiy.marathon.apple.logparser.TestEventProducer {
+class DeviceFailureParser : TestEventProducer {
     private val patterns = listOf(
         "Failed to install or launch the test runner",
         "Software caused connection abort",
--- a/vendor/vendor-apple/base/src/main/kotlin/com/malinskiy/marathon/apple/logparser/parser/TestRunProgressParser.kt
+++ b/vendor/vendor-apple/base/src/main/kotlin/com/malinskiy/marathon/apple/logparser/parser/TestRunProgressParser.kt
@@ -30,7 +30,10 @@ class TestRunProgressParser(
      */
     val FAILING_TEST_MATCHER = "(/.+:\\d+):\\serror:\\s[\\+\\-]\\[(.*)\\s(.*)\\]\\s:(\\s.*)".toRegex()
 
-    val TIMEOUT_TEST_MATCHER = """Test Case '-\[(.+) (.+)]' exceeded execution time allowance of (\d+) minutes\. The test may have hung.*""".toRegex()
+    /**
+     * Timeout case from https://developer.apple.com/documentation/xctest/xctestcase/3526064-executiontimeallowance
+     */
+    val TIMEOUT_TEST_MATCHER = """Test Case '-\[(.+) (.+)]' exceeded execution time allowance of (\d+) minutes*\. The test may have hung.*""".toRegex()
     
     private var failingTestLine: String? = null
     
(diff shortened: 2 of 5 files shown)


Malinskiy:
Thanks for sending this.
I think it's a good addition to handling edge cases for Apple

Can you please double check if timeout case produces 0 value in the [TemporalTestResult](https://github.com/MarathonLabs/marathon/blob/8b5087fc2d8e6fe982e981ec79188c08bd8cf2c7/core/src/main/kotlin/com/malinskiy/marathon/execution/result/TemporalTestResult.kt#L8) please?
There is also a piece of logic [here](https://github.com/MarathonLabs/marathon/blob/8b5087fc2d8e6fe982e981ec79188c08bd8cf2c7/core/src/main/kotlin/com/malinskiy/marathon/execution/result/TestRunResultsAccumulator.kt#L183) that is supposed to fill in those 0's, I wonder why it's not invoked, assuming this is the source your issue with negative durations.

msdeleyto:
> Thanks for sending this. I think it's a good addition to handling edge cases for Apple
> 
> Can you please double check if timeout case produces 0 value in the [TemporalTestResult](https://github.com/MarathonLabs/marathon/blob/8b5087fc2d8e6fe982e981ec79188c08bd8cf2c7/core/src/main/kotlin/com/malinskiy/marathon/execution/result/TemporalTestResult.kt#L8) please? There is also a piece of logic [here](https://github.com/MarathonLabs/marathon/blob/8b5087fc2d8e6fe982e981ec79188c08bd8cf2c7/core/src/main/kotlin/com/malinskiy/marathon/execution/result/TestRunResultsAccumulator.kt#L183) that is supposed to fill in those 0's, I wonder why it's not invoked, assuming this is the source your issue with negative durations.

Ok, I have been running some tests and debugging the solution. I did not finish but wanted to register the progress here and let you know the status.

### TemporalTestResult

Yes, this use case produces a value of 0 for the `endTime` property.

The problem seems to be in the `TestRunProgressParser`. Without the check I added it does not detect the test finished. This results into the `AppleSimulatorDevice` not detecting the test finished event (it does not exist).

```kt
...
for (line in session.stdout) {
    testEventProducer.process(line)?.let {
        send(it)
    } 
    lineListeners.forEach { it.onLine(line) }
}
...
``` 

The test is then detected as `uncompleted` and the `TestsResultListener` process this case here:
```kt
if (uncompleted.isNotEmpty()) {
    uncompleted.forEach {
        logger.warn { "uncompleted = ${it.test.toTestName()}, ${device.serialNumber}" }
    }
}
```
There I could print the `endTime` property (being 0).

### Fill `endTime` strategy

Here is where I need to keep testing. I log that function and is not being invoked but I could not determine why that is happening.

Regarding the timeout from Apple, more information in [Apple docs](https://developer.apple.com/documentation/xctest/xctestcase/3526064-executiontimeallowance). Basically it is specified in seconds and it is round up to minutes so we can expect to use (in regex) `minutes*`.

### How to reproduce

- Create a XCUITest with a delay of 61 seconds.
- For the vendor configuration, enable the timeout for the minimum (1 minute):
  ```yml
    xcodebuildTestArgs:
      "-test-timeouts-enabled": "YES"
      "-maximum-test-execution-time-allowance": "60"
  ```
- The test should be reported as non completed and the log:
  ```kt
  "Test $pkg.$clazz.$method finished with result <$result> after $duration seconds"
  ```
  should be missing.

msdeleyto:
After checking again, it seems the `fillEndTime` function is not used because it is a non completed test

msdeleyto:
@Malinskiy can we proceed with the PR?

Malinskiy (approved):

Malinskiy (approved):

Malinskiy:
FYI looking at the XCTestCore string templates, the timeout value might not be minutes, so the regex will fail if it's seconds for example

msdeleyto:
> FYI looking at the XCTestCore string templates, the timeout value might not be minutes, so the regex will fail if it's seconds for example

According to their docs I understood it could be only minute(s) because it is rounded up to the nearest minute. In any case something like this should do the trick:
```
...exceeded execution time allowance of (\d+) (minute[s]*|seconds)\. The test may have hung...
```

Malinskiy:
FYI I've implemented parsing of seconds, minutes and hours just in case. It's part of the latest release

PR #996, opened 2025-04-15
AnnotationDataFilter: support annotations with stringArray type values

koctuk999:
https://github.com/MarathonLabs/marathon/issues/994


Malinskiy:
@koctuk999 please add tests with string array value as well as some other type to check how this code behaves

koctuk999:
> @koctuk999 please add tests with string array value as well as some other type to check how this code behaves

added tests

Malinskiy:
FYI this might break in the future, but since it's an undocumented filter I'm ok with that.
There is a use-case for all of the values to match the regex, not any. If this comes up - will need to rework this

Malinskiy (approved):

PR #1000, opened 2025-04-29
(iOS) Fix empty failure tag on JUnit XML report when unknown class

msdeleyto:
Proposal to fix the issue reported [here](https://github.com/MarathonLabs/marathon/issues/999).

When the xcode execution log prints a test failure with an unknown class, the parser does not detect the error trace with the current regex expression, expecting a proper class path (`/path/to/class`):
```log
...
<DebugLogPrinter> <unknown>:0: error: -[TestTarget.Tests testAccessibility] : Element has no description
<DebugLogPrinter>     t =     4.79s Find the Any (Descendant with identity binding to AX element pid: 62723, elementOrHash.elementID: 105553170498336.190)
<DebugLogPrinter>     t =     4.80s Checking existence of `"empty state view" StaticText`
<DebugLogPrinter>     t =     4.81s Find the "empty state view" StaticText
...
```

which results into the following failure tag:
```xml
    ...
    <testcase name="testAccessibility" time="13.387" classname="TestTarget.Tests">
      <failure></failure>
    </testcase>
    ...
```

For now I do not understand why the log says "unknown" class. However, this change would provide some information on the JUnit XML report about the stacktrace, as a first step to tackle the issue.

Note 1: When multiline stacktrace the parser only captures the first line.
Note 2: as an additional change the related test could be parameterized to simplify the addition of new tests / maintenance of the test.

### DIFF
--- a/vendor/vendor-apple/base/src/test/kotlin/com/malinskiy/marathon/apple/logparser/parser/TestRunProgressParserTest.kt
+++ b/vendor/vendor-apple/base/src/test/kotlin/com/malinskiy/marathon/apple/logparser/parser/TestRunProgressParserTest.kt
@@ -5,12 +5,21 @@ import assertk.assertions.isEqualTo
 import com.malinskiy.marathon.apple.test.TestEvent
 import com.malinskiy.marathon.time.Timer
 import org.junit.jupiter.api.BeforeEach
-import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.extension.ExtensionContext
+import org.junit.jupiter.params.ParameterizedTest
+import org.junit.jupiter.params.provider.Arguments
+import org.junit.jupiter.params.provider.ArgumentsProvider
+import org.junit.jupiter.params.provider.ArgumentsSource
+import org.junit.runner.RunWith
+import org.junit.runners.Parameterized
 import org.mockito.Mockito.reset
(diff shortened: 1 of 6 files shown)


Malinskiy:
`<unknown>:0: error` indicates a failure in some compiler-generated code that doesn't have a source, hence it has nothing to point to. Approved.

vendor/vendor-apple/base/src/test/kotlin/com/malinskiy/marathon/apple/logparser/parser/TestRunProgressParserTest.kt
```diff
@@ -21,114 +29,31 @@ class TestRunProgressParserTest {
...
-        val events = mutableListOf<TestEvent>()
-        javaClass.getResourceAsStream("/fixtures/test_output/success_multiple_0.log.input").bufferedReader().use {
-            it.lines().forEach { line ->
-                parser.process(line)?.let {
-                    events.addAll(it)
-                }
-            }
+    data class TestData(val caseName: String, val targetName: String)
+
+    class TestDataProvider : ArgumentsProvider {
```
Malinskiy, line 34: Appreaciate the time to redo this properly

Malinskiy (approved):

PR #1015, opened 2026-04-27
Introduce allowEmptyTestSuites parameter

smplio:
Introduce allowEmptyTestSuites parameter to not fail if no tests were discovered to run.
This is very useful when trying to run tests on a whole multi-module android project using marathon-gradle-plugin


dhabensky:
+1, looks useful for multi-module and multi-app projects

